PR #5, opened 2019-09-03
Implement reqwest features: bearer_auth and form

dbdr:
I'm interested in making it easier to use leaner HTTP clients like attohttpc instead of the huge dependencies pulled by reqwest, especially when async execution is not required. Attohttpc seems quite good for this since the API is very close to reqwest (by design?). While experimenting with this, I stumbled upon these two missing features:

RequestBuilder::bearer_auth, which is just a convenience method to set an Authorization header of a specific form.

RequestBuilder::form, which is quite similar to RequestBuilder::json, so I tried to keep my implementation as close to it as possible, including making it an optional feature.

I hope these additions will reduce the friction of switching from reqwset to attohttpc. Please review the code and let me know if you suggest any modifications.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ libflate = { version = "0.1", optional = true }
 native-tls = { version = "0.2", optional = true }
 serde = { version = "1", optional = true }
 serde_json = { version = "1", optional = true }
+serde_urlencoded = { version = "0.6", optional = true }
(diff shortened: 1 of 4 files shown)


sbstp:
Thanks, this is great. I'll release new version.

sbstp:
It's in 0.4.4

PR #8, opened 2019-11-06
Make prepared requests reusable

adamreichold:
`PreparedRequest::send` currently consumes the request even though this is not really necessary. This changes the signature so that a single prepared request can be reused multiple times.

To make this possible the host header has to be updated each time, but this actually saves one clone of the URL with the cost of delaying some of the errors triggered by `PreparedRequest::set_host`.


sbstp:
Hey. Thanks for your pull requests! I wanted to review & merge them last night but I fell asleep. I'll do it ASAP.

PR #9, opened 2019-11-06
Enable generic body types to support borrowed bodies

adamreichold:
Making the request body generic with an `AsRef<[u8]>` trait bound keeps the implementation basically the same while allowing borrowed bodies which for example enables the user to avoid copies when multiple requests with the same body need to be made.

This happens for example when a request has to be retried against different hosts which is necessary with some distributed services.

The two major downsides I currently see here are:
* Possible code bloat as the methods of `RequestBuilder` and `PreparedRequest` are now monomorphized.
* Surprise type changes as the top-level functions always create `RequestBuilder<Vec<u8>>` which the modifier would then change to some other `RequestBuilder<T>`. 

As an alternative, a public setter for the URL of `PreparedRequest` as a follow-up to #8 would serve the same purposes with much less API churn but would also be strictly less general. For example using this PR, multiple threads could make concurrent requests using the same body behind an `Arc<Vec<u8>>` or the body could be a memory-mapped file opened using the `memmap` crate.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -82,7 +82,7 @@ impl RequestBuilder {
     where
         U: AsRef<str>,
     {
-        Self::with_body(method, base_url, [])
+        Self::try_new(method, base_url).expect("invalid url or method")
     }
 
     /// Try to create a new `RequestBuilder`.
@@ -92,32 +92,6 @@ impl RequestBuilder {
     pub fn try_new<U>(method: Method, base_url: U) -> Result<Self>
     where
         U: AsRef<str>,
-    {
-        Self::try_with_body(method, base_url, [])
-    }
-}
-
-impl<B> RequestBuilder<B> {
-    /// Create a `Request` with the base URL, the given method and the given body.
-    ///
-    /// # Panics
-    /// Panics if the base url is invalid or if the method is CONNECT.
-    pub fn with_body<U>(method: Method, base_url: U, body: B) -> Self
-    where
-        U: AsRef<str>,
-        B: AsRef<[u8]>,
-    {
-        Self::try_with_body(method, base_url, body).expect("invalid url or method")
-    }
-
-    /// Try to create a `RequestBuilder` with the given body.
-    ///
-    /// If the base URL is invalid, an error is returned.
-    /// If the method is CONNECT, an error is also returned. CONNECT is not yet supported.
-    pub fn try_with_body<U>(method: Method, base_url: U, body: B) -> Result<Self>
-    where
-        U: AsRef<str>,
-        B: AsRef<[u8]>,
     {
         let url = Url::parse(base_url.as_ref()).map_err(|_| ErrorKind::InvalidBaseUrl)?;
 
@@ -129,7 +103,7 @@ impl<B> RequestBuilder<B> {
             url,
             method,
             headers: HeaderMap::new(),
-            body,
+            body: [],
             max_redirections: 5,
             follow_redirects: true,
             #[cfg(feature = "charsets")]
@@ -138,7 +112,9 @@ impl<B> RequestBuilder<B> {
             allow_compression: true,
         })
     }
+}
 
+impl<B> RequestBuilder<B> {
     /// Associate a query string parameter to the given value.
     ///
     /// The same key can be used multiple times.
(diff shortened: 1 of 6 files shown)


sbstp:
This change makes sense, it's best to avoid unnecessary allocation when possible. I will release this as 0.5 to signal a breaking change. There is a merge conflict though, can you fix that?

adamreichold:
> There is a merge conflict though, can you fix that?

Will do.

sbstp:
It would also be nice to remove the allocation in the `text` and `bytes` methods. I don't think it would be easy to avoid allocating for the JSON method though.

adamreichold:
> It would also be nice to remove the allocation in the `text` and `bytes` methods.

I did not do that because I wanted to avoid changing the API too much, but if that would be alright with you, I will reduce these to take `AsRef<str>` and `AsRef<[u8]>` instead of taking ownership.

> I don't think it would be easy to avoid allocating for the JSON method though.

Not without supporting chunked encoding, i.e. one could hold on to a `T: Serialize` body and serialize it directly onto the HTTP connection but the length would not be known upfront.

adamreichold:
If we are willing to make more intrusive changes, what you do think of changing the default body type to `[u8; 0]`, i.e. a zero-sized type, to avoid moving the three pointers of an empty `Vec<u8>` around?

adamreichold:
> Will do.

Done.

> I did not do that because I wanted to avoid changing the API too much, but if that would be alright with you, I will reduce these to take AsRef<str> and AsRef<[u8]> instead of taking ownership.

Added another commit doing that.

adamreichold:
> If we are willing to make more intrusive changes, what you do think of changing the default body type to [u8; 0], i.e. a zero-sized type, to avoid moving the three pointers of an empty Vec<u8> around?

Added a commit that does. Please feel free to drop it from the PR if you do not like the change.

src/request.rs
```diff
@@ -194,71 +220,94 @@ impl RequestBuilder {
...
+                self.0.as_ref().as_bytes()
+            }
+        }
+
         self.headers
             .entry(http::header::CONTENT_TYPE)
             .unwrap()
             .or_insert(HeaderValue::from_static("text/plain; charset=utf-8"));
-        self
+        self.body(Text(body))
```
sbstp: Doesn'tt `body.as_ref()` work here? `str` implements `AsRef<u8>`

sbstp: I checked out the code and I see now that you need to own the `impl AsRef<str>` where `.as_ref()` simply borrows it and that doesn't work. Neat work around.

sbstp:
I did not realize at first that using the builder pattern could allow a sneaky update of the generic type but it seems like this is what is happening with the `body` function, that's a neat trick. I think the `body` function should be private though. It could lead to unexpected behavior because `body` does not set the content-type. If the user uses `text` or `bytes` we can set the content-type header properly. So in order to prevent misuse, `body` should be private.

sbstp:
Other than my previous comment this looks really good. I don't have a problem with the empty u8 array. It's a really clever way of offering a clean API :+1:

sbstp:
> Surprise type changes as the top-level functions always create RequestBuilder<Vec<u8>> which the modifier would then change to some other RequestBuilder<T>.
I think we could potentially hide this with a virtual dispatch if we have complaints. It would also reduce the amount of code generation. I'm not really worried by a virtual dispatch in IO bound code.

adamreichold:
> So in order to prevent misuse, body should be private.

Added a commit that does this.

> I think we could potentially hide this with a virtual dispatch if we have complaints. It would also reduce the amount of code generation.

I tried that but it gets messy because you have to abstract over whether `RequestBuilder` contains `Box<dyn AsRef<[u8]>>` or `&'a dyn AsRef<[u8]>`, i.e. use an `enum` with both variants  or resort to nested trait bounds like `AsRef<dyn AsRef<[u8]>>` which could still end up with many instantiations.

> I'm not really worried by a virtual dispatch in IO bound code.

I agree.

sbstp:
Ok. Good enough for now. Thanks a lot, I will release 0.5.0 right away.

sbstp:
It's published.

PR #12, opened 2019-11-11
WIP: Use Cow to hold request body

sbstp:


adamreichold:
The problem I see with this approach is that the lifetime `'b` now infects everything, i.e. even a fully owned body has this meaning that I cannot own/box a `PreparedRequest` besides trying to force that lifetime to `'static` which seems a bit unergonomic to me. (I am not sure if it actually works at all to be honest.)

It is also still generic code as lifetimes are type parameters. And it rules out other types that can usefully provide `AsRef<[u8]>`, e.g. `Arc<Vec<u8>>` which I used several times when I shared bodies between threads.

So I think having a general `Body` trait to really avoid allocation in the JSON case would be more useful, but that means supporting chunked transfer encoding as the size is not known upfront.

src/request.rs
```diff
@@ -200,75 +202,56 @@ impl<B> RequestBuilder<B> {
...
+        self.body = Cow::Borrowed(body);
+        self
     }
 
     /// Set the body of this request to be the JSON representation of the given object.
     ///
     /// If the `Content-Type` header is unset, it will be set to `application/json` and the charset to UTF-8.
     #[cfg(feature = "json")]
-    pub fn json<T: serde::Serialize>(mut self, value: &T) -> Result<RequestBuilder<impl AsRef<[u8]>>> {
+    pub fn json<T: serde::Serialize>(mut self, value: &T) -> Result<Self> {
```
adamreichold: Since this is an owned variant, shouldn't it erase the lifetime and return `RequestBuilder<'static>`?

PR #14, opened 2019-11-12
Use ToString instead of Display as it is strictly more general

adamreichold:
Use `value.to_string()` instead of `format!("{}", value)` as `ToString` is strictly more general due to the blanket implementation based on `Display`.

This also slightly generalizes `RequestBuilder::params` to take any iterable and makes its test a doc test to extend its usefulness for library users.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,8 +1,8 @@
 #![allow(dead_code)]
 #![allow(clippy::write_with_newline)]
+use std::borrow::Borrow;
 use std::borrow::Cow;
 use std::convert::From;
-use std::fmt::Display;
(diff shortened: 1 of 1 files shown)


sbstp:
There's a conflict again :o . I started experimenting with a simpler design for the body parameters if you want to check #12 . I think we could borrow the body fully to avoid allocation, even in the case of json with something like:
```rust
enum Body<'b, B> where B: Serialize {
  Bytes(&'b [u8]),
  Json(&'b B),
  Form(&'b B),
}
```

adamreichold:
> There's a conflict again :o .

Resolved.

PR #15, opened 2019-11-22
Support HTTP basic auth via OpenSSL Base64 support

adamreichold:
OpenSSL provides TLS support on Unix platforms, but also includes Base64 encoding which allows us to support HTTP basic authentication without additional dependencies.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -206,6 +206,24 @@ impl<B> RequestBuilder<B> {
         Ok(self)
     }
 
+    /// Enable HTTP basic authentication.
+    ///
+    /// This is available only on Linux and when TLS support is enabled.
+    #[cfg(all(
+        feature = "tls",
+        not(any(target_os = "windows", target_os = "macos", target_os = "ios"))
+    ))]
+    pub fn basic_auth(self, username: impl std::fmt::Display, password: Option<impl std::fmt::Display>) -> Self {
+        let auth = match password {
+            Some(password) => format!("{}:{}", username, password),
+            None => format!("{}:", username),
+        };
+        self.header(
+            http::header::AUTHORIZATION,
+            format!("Basic {}", openssl::base64::encode_block(auth.as_bytes())),
+        )
+    }
+
     /// Enable HTTP bearer authentication.
     pub fn bearer_auth(self, token: impl Into<String>) -> Self {
         self.header(http::header::AUTHORIZATION, format!("Bearer {}", token.into()))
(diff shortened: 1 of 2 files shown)


sbstp:
I think the OpenSSL library was released with your changes.

Could you offer a fallback that uses a base64 library when the TLS feature is off or when it's not linux? The docs.rs show the Linux version usually and they will probably get a surprise if they try out the basic auth on Windows or Darwin.

adamreichold:
> I think the OpenSSL library was released with your changes.

Thanks for the heads-up! Updated the PR to remove patching in the Git dependencies.

> Could you offer a fallback that uses a base64 library when the TLS feature is off or when it's not linux? The docs.rs show the Linux version usually and they will probably get a surprise if they try out the basic auth on Windows or Darwin.

So you would be willing to take an additional dependency on e.g. the `base64` crate for that or do the other `native-tls` backends provide Base64 coding as well? If we do take an additional dependency it would probably not be worth it to go through the OpenSSL API at all and I could just do basic auth as a separate feature flag.

But honestly speaking, my sole motivation for the OpenSSL changes were to avoid additional dependencies here as a minimal set of dependencies seems to be one the main selling points of this crate.

I could also use the additional dependency only on Windows/Darwin, but that would feel quite complicated? Or we could just extend the doc comment to indicate that this requires the TLS feature and a Unix platform?

adamreichold:
> Or we could just extend the doc comment to indicate that this requires the TLS feature and a Unix platform?

Added this as the minimum one can do about this.

adamreichold:
Seems like e.g. schannel handles Base64 coding internally, but does not expose API to do it.

sbstp:
I was thinking of just using the `base64` crate only on Windows and Darwin. If you really don't want to add dependencies, then you can just comment the fact that the function only exists with the TLS feature on Linux.

src/request.rs
```diff
@@ -3,6 +3,8 @@
 use std::borrow::Borrow;
 use std::borrow::Cow;
 use std::convert::From;
+#[cfg(all(feature = "tls", unix))]
```
sbstp: Does darwin not count as a unix in this case?

sbstp: https://github.com/sfackler/rust-native-tls/blob/master/Cargo.toml#L23

adamreichold: Indeed! Fixed.

sbstp:
Thanks!

sbstp:
It's out in 0.5.2

adamreichold:
> It's out in 0.5.2

Thanks! I am not completely sure, but we did bump `url` from 1.7 to 2.1 so this seems to have been a breaking change there which would imply a breaking here as we expose the `Url` in our API. So shouldn't that have been 0.6.0?

sbstp:
I think you're be right. I'll yank 0.5.2 and release 0.6.0

PR #17, opened 2019-12-05
Update http dependency to current stable version

adamreichold:
The `http` recently saw a [new stable release](https://seanmonstar.com/post/189439210962/http-v02) and the ecosystem is expected to move there eventually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ keywords = ["http", "client", "tls"]
 categories = ["network-programming", "web-programming", "web-programming::http-client"]
 
 [dependencies]
-http = "0.1"
+http = "0.2"
 log = "0.4"
 url = "2"
 encoding_rs = { version = "0.8", optional = true }
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,4 @@
+use std::convert::Infallible;
 use std::error::Error as StdError;
 use std::fmt::{self, Display};
 use std::io;
@@ -147,6 +148,12 @@ impl StdError for Error {
     }
 }
 
+impl From<Infallible> for Error {
+    fn from(_err: Infallible) -> Error {
+        unreachable!()
+    }
+}
+
 impl From<io::Error> for Error {
     fn from(err: io::Error) -> Error {
         Error(Box::new(ErrorKind::Io(err)))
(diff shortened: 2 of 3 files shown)


sbstp:
Ah sweet `TryInto` is in stable. Thanks again for this!

By the way, the library was [mentionned in a blog post](https://stjepang.github.io/2019/12/04/blocking-inside-async-code.html) by one of the async std authors, it seems like it's gaining some legitimacy.

adamreichold:
> By the way, the library was [mentionned in a blog post](https://stjepang.github.io/2019/12/04/blocking-inside-async-code.html) by one of the async std authors, it seems like it's gaining some legitimacy.

Thanks for the link. I have to admit I am somewhat wary of the content though as it seems another example of an infatuation with even-driven code. While I am glad that async/await is stable we started migrating to Tokio 0.2 at $DAYJOB this week, I fear some folks have lost the perspective that the cooperative multitasking of event-driven runtimes is one way to handle concurrency and parallelism among many. Which I fear is especially dangerous now that the much improved ergonomics that come with async/await make it that much easier to forget the issues related to this particular programming model.

sbstp:
I've released this in `0.7.0`. 

I feel like async is often using excessive force to solve something that might not even be a problem, not that many people need to scale to a level where async runtimes will make a difference.

I'm glad to hear that your workplace uses Rust. I hope that I'll be able to introduce it at my workplace eventually. Currently we use Go. It's pretty fun to work with because all the async details are hidden away. They could switch to another concurrency model and it wouldn't break any code. The code doesn't get polluted by a bunch of async/await keywords. Still has a bunch of issues with the lack of generics and what not though, but I like the idea of what they're trying to do.

Also Go is much closer to C than Rust in terms of safety. It's really easy to introduce nasty bugs.

sbstp:
Thanks again!

adamreichold:
> I've released this in `0.7.0`.

Thanks!

> I'm glad to hear that your workplace uses Rust. I hope that I'll be able to introduce it at my workplace eventually. Currently we use Go. It's pretty fun to work with because all the async details are hidden away. They could switch to another concurrency model and it wouldn't break any code. The code doesn't get polluted by a bunch of async/await keywords. Still has a bunch of issues with the lack of generics and what not though, but I like the idea of what they're trying to do.

We use a mixture of Rust, Go, C and Python. Rust is definitely replacing C for low-level network programming for us. We do have some issues with more complex domain logic in Go slowly becoming unmaintainable, but that might be as much our problem as it is Go's. And we are definitely porting some performance-critical code paths from Go to Rust as well as we had serious GC-related trouble with throughput-oriented database processing in Go. I do agree that the ergonomics of network programming using Go are pretty hard to beat if one is able to pay the price for the language runtime the makes it possible.

PR #20, opened 2019-12-28
Implement Happy Eyeballs

sbstp:


src/happy.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    res
+}
+
+/// aaa
+pub fn connect<A>(addrs: A) -> io::Result<TcpStream>
+where
+    A: ToSocketAddrs,
+{
+    let addrs: Vec<_> = addrs.to_socket_addrs()?.collect();
```
adamreichold: `intertwine` seem quite general for this particular case and by rewriting it so that it handles a single `ToSocketAddrs` input directly seems like it could avoid the first `collect` call? IIRC the RFC also suggest to prefer IPv6 whereas this would spawn the connection to the first IPv4 address first?

src/happy.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    thread::spawn(move || {
+        for addr in order {
+            if done.load(Ordering::SeqCst) {
+                break;
+            }
+
+            let tx = tx.clone();
+            thread::spawn(move || {
+                match TcpStream::connect_timeout(&addr, Duration::from_secs(10)) {
```
adamreichold: Wouldn't it make sense to make this timeout a parameter of `happy::connect`?

src/happy.rs
```diff
@@ -0,0 +1,84 @@
...
+pub fn connect<A>(addrs: A) -> io::Result<TcpStream>
+where
+    A: ToSocketAddrs,
+{
+    let addrs: Vec<_> = addrs.to_socket_addrs()?.collect();
+    let ipv4 = addrs.iter().filter(|a| a.is_ipv4()).cloned();
+    let ipv6 = addrs.iter().filter(|a| a.is_ipv6()).cloned();
+    let order = intertwine(ipv4, ipv6);
+
+    let done = Arc::new(AtomicBool::new(false));
```
adamreichold: I wonder if the additional synchronization via `done` and the intermediate thread to spawn the connection threads is really helpful. Couldn't we remove this and replace the `thread::sleep` with a `recv_timeout` call on `rx`?

src/happy.rs
```diff
@@ -0,0 +1,84 @@
...
+                };
+            });
+
+            thread::sleep(Duration::from_millis(200));
+        }
+    });
+
+    let res = match rx.recv() {
+        Ok(sock) => Ok(sock),
+        Err(_) => Err(io::ErrorKind::ConnectionRefused.into()),
```
adamreichold: Maybe the threads should send both sockets and errors on this channel so that if no connection could be established, we could yield the first error here instead?

src/happy.rs
```diff
@@ -0,0 +1,84 @@
...
+use std::iter::FusedIterator;
+use std::net::TcpStream;
+use std::net::ToSocketAddrs;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::mpsc::channel;
+use std::sync::Arc;
+use std::thread;
+use std::time::{Duration, Instant};
+
+fn intertwine<T, A, B>(mut ita: A, mut itb: B) -> Vec<T>
```
adamreichold: It is definitely a nit, but at work we usually begin a module with the public API and have the implementation details at the bottom so it is easier for an unfamiliar reader to get the gist of a module without being swamped by helper types and functions.

src/happy.rs
sbstp: > intertwine seem quite general for this particular case and by rewriting it so that it handles a single ToSocketAddrs input directly seems like it could avoid the first collect call?

I went with a collect here so I could create two iterators on the same list of addresses. `to_socket_addrs` returns an iterator, I put then in a vec to avoid having to call `to_socket_addrs` twice which would perform DNS resolution twice. 

> IIRC the RFC also suggest to prefer IPv6 whereas this would spawn the connection to the first IPv4 address first?

The RFC says that you can sort addresses IPv4 first if you like. I figured that in the typical use case of the library, which is server apps and desktop apps, IPv4 would probably work better. IPv6 is popular in mobile networks but I don't see this library being used in mobile apps.

src/happy.rs
sbstp: Yes, this is a WIP, the timeout will be a parameter of the function.

src/happy.rs
sbstp: Yes that's a good idea. It would avoid spawning a thread for the loop as well. I've also been looking at threadpool libraries, I'm not sure if the performance cost of spawning threads like this is big enough to warrant using a thread pool library.

src/happy.rs
sbstp: That's a good idea. I wasn't quite sure which synthetic IO error to produce here, so this would fix that issue.

src/happy.rs
sbstp: This is a WIP so it's a bit messy. I might move this to a util module or something.

src/happy.rs
adamreichold: Personally, I would not go for a thread pool here to keep things simple. It might make sense to limit the number of threads spawned though, e.g. try to connect to at most `N` addresses in parallel by effectively throttling the loop.

sbstp: I did a few experiments, nothing very thorough but I usually don't see a ton of addresses attached to a domain, so I think that for now we can leave it like this and limit the number of threads if problems occur.

src/happy.rs
adamreichold: > I put then in a vec to avoid having to call to_socket_addrs twice which would perform DNS resolution twice.

You are right, there is no avoiding the first collection. Maybe prefer `copied` over `cloned` to make it explicit that this is cheap? One could even write this as an iterator adaptor I guess to avoid the second collection, but I am not sure if it makes this clearer, e.g.
```
fn intertwine<T, A, B>(mut ita: A, mut itb: B) -> impl Iterator<Item = T>
where
    A: FusedIterator<Item = T>,
    B: FusedIterator<Item = T>,
{
    let mut stashed = None;

    std::iter::from_fn(move || {
        if stashed.is_some() {
            return stashed.take();
        }

        match (ita.next(), itb.next()) {
            (Some(a), Some(b)) => {
                stashed = Some(b);
                Some(a)
            }
            (Some(a), None) => Some(a),
            (None, Some(b)) => Some(b),
            (None, None) => None,
        }
    })
}
```

> IPv6 is popular in mobile networks but I don't see this library being used in mobile apps.

Personally, I would like it if IPv6 was preferred to generally push everybody to use IPv6 wherever it becomes available.

src/happy.rs
```diff
@@ -0,0 +1,95 @@
...
+            }
+            Err(RecvTimeoutError::Disconnected) => {
+                unreachable!();
+            }
+        }
+    }
+
+    debug!(
+        "could not connect to any address, took {}ms",
+        (Instant::now() - start).as_millis()
```
adamreichold: `Instant::elapsed`

src/happy.rs
```diff
@@ -0,0 +1,95 @@
...
+                let end = Instant::now();
+                let span = end - start;
+                debug!("success, took {}ms", span.as_millis());
+
+                return Ok(sock);
+            }
+            Ok(Err(err)) => {
+                debug!("connection error: {} addr={}", err, addr);
+
+                last_err = Some(err);
```
adamreichold: I think the first error might be more appropriate if we consider the this the highest priority address? But that is admittedly rather unscientific.

src/happy.rs
adamreichold: Agreed.

src/happy.rs
```diff
@@ -0,0 +1,95 @@
...
+            }
+            Err(RecvTimeoutError::Timeout) => {
+                continue;
```
adamreichold: Don't we still need a fully blocking `recv` after the loop? E.g. we spawn two threads but nothing connects in 400ms and hence we bail out, but the first thread could just succeed after 500ms? So I think a second call to `recv` is necessary.

src/happy.rs
adamreichold: But due to the threading, there are no guarantees anyway. So this is really moot. Sorry for the noise.

src/happy.rs
sbstp: Yeah you're right. I think I have to loop after the for loop in case they all timed out.

src/happy.rs
sbstp: Fixed

src/happy.rs
sbstp: Fixed

src/happy.rs
sbstp: Ok, it's not a big deal to me. My ISP does not support IPv6 right now, so I tried it out configured with IPv6 first and it doesn't seem to delay me much. The IPv6 connections fails quickly and IPv4 takes over.

src/happy.rs
sbstp: I added a second loop to ensure we catch everything, let me know what you think.

src/happy.rs
```diff
@@ -0,0 +1,128 @@
...
+/// to a domain which is available in both IPv4 and IPv6. Connection attempts are raced
+/// against each other and the first to connect successfully wins the race.
+///
+/// If the timeout is not provided, a default timeout of 10 seconds is used.
+pub fn connect<A>(addrs: A, timeout: impl Into<Option<Duration>>) -> io::Result<TcpStream>
+where
+    A: ToSocketAddrs,
+{
+    let timeout = timeout.into().unwrap_or(DEFAULT_CONNECTION_TIMEOUT);
+    let addrs: Vec<_> = addrs.to_socket_addrs()?.collect();
```
sbstp: Could add a fast path if there's only a single address returned.

src/happy.rs
adamreichold: I think you could avoid `deadline` und just use `recv` instead of `recv_timeout` as the threads will just hang up until they timed out you will break the loop due to the `Err(_)` match arm.

src/happy.rs
adamreichold: Thanks! (My ISP supports IPv6 as well, so things make not seem as bleak as they used to w.r.t. IPv6 deployment.)

(I think the iterator adaptor version of `intertwine` from above would still be nice)

src/happy.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    if addrs.len() == 1 {
```
adamreichold: I think this could be a slice pattern like `if let [addr] = addrs.as_slice()` to check and unpack in one go.

src/happy.rs
sbstp: Yeah you're absolutely right. At first I forgot to drop the tx so the disconnection was not happening. But I added the drop later and now it does occur so the timeout is not necessary.

adamreichold:
LGTM to me now. Just two style nits: You might want to `use std::iter;` to be consistent with how you use e.g. `std::thread` and we might want to return `Other` instead of `ConnectionRefused´ as that one has a rather specific meaning (received a RST from the other end) and we should only end up there if no addresses were returned by `to_socket_addrs` at all.

sbstp:
I think that the `to_socket_addrs` call will return an error if no DNS entries are found, but just to make the code exhaustive I'll return `Other`.

PR #23, opened 2019-12-29
Make BodyReader a BufRead implementation

adamreichold:
To do so, we replace `LengthReader` by the standard libraries' `io::Take` and turn `ChunkedReader` into a `BufRead` implementation as well by buffering exactly one chunk at a time. We can thereby remove the extra buffering done behind the scenes by the `flate2` crate.


src/parsing/chunked_reader.rs
```diff
@@ -34,95 +32,64 @@ where
...
-        }
+    fn fill_buf(&mut self) -> io::Result<&[u8]> {
+        if self.buffer.len() == self.consumed && !self.reached_eof {
+            let chunk_size = self.read_chunk_size()?;
 
-        // If we have a length of 0 for a chunk, we've reached EOF.
-        if self.length == 0 {
-            return Ok(0);
-        }
+            self.buffer.resize(chunk_size + 2, 0);
```
sbstp: I'm a bit concerned by this, if a server sends large chunks it will cause the library to allocate a lot of memory. I think we can still implement BufRead without buffering the whole chunk at once.

src/parsing/chunked_reader.rs
```diff
@@ -34,95 +32,64 @@ where
...
+            self.inner.read_exact(&mut self.buffer)?;
 
-        // We read the smallest amount between the given buffer's length and the remaining bytes' length.
-        let count = cmp::min(self.remaining(), buf.len() as u64) as usize;
+            if !self.buffer.ends_with(b"\r\n") {
```
sbstp: The current way accepts both `\n` and `\r\n`. I like the saying "be conservative in what you do, be liberal in what you accept from others". I don't think it's easy to accept both ways with a read_exact though.

src/parsing/chunked_reader.rs
adamreichold: Yes, if we track how much of the chunk we still need to read the next time we have to refill the buffer and switch on whether we are mid-chunk or not.

But this does raise two question: What is a sensible amount of buffering? (For example the standard library internally goes for 8 kB buffers when e.g. `BufReader::new` is used.) And do want to try to always fill this buffer even if it means parsing multiple chunks which complicate things further?

Another approach might be to just limit the chunk size to e.g. 16 bit and hence 64 kB chunks.

Also note that we already have quite a few such issues in the code base, e.g. `read_chunk_size` will happily read a 3 GB buffer of digits until terminated by `\n` before it actually tries to parse this into number. So if we are not going to fix them all by limiting buffers everywhere, it would be unfortunate to loose the simplicity IMHO of buffering one chunk at a time.

src/parsing/chunked_reader.rs
adamreichold: Yes, that is certainly a change in accepted data. But I think it can be easily fixing with a bit of ceremony as the underlying reader is already buffered, i.e. I can `read_exact` only `chunk_size + 1` and then read another character `\r\n` if the last one was `\r`.

src/parsing/chunked_reader.rs
sbstp: > And do want to try to always fill this buffer even if it means parsing multiple chunks which complicate things further?

I think we could just return a shorter buffer when the chunk is smaller the entire buffer.

> Also note that we already have quite a few such issues in the code base, e.g. read_chunk_size will happily read a 3 GB buffer of digits until terminated by \n before it actually tries to parse this into number. 

That's very true, I think it's something we might want to address eventually. It's probably more important for a server to limit the sizes of the data received because bad actors might try to DDOS it. I don't think that it's likely that a malicious server would try to break our library by announcing large chunks. I think that it's more likely that a server will unintentionally return large chunks and cause the library to use a lot of memory. I think we should put a hard cap on the buffer size at like 32 or 64 KiB.

src/parsing/compressed_reader.rs
```diff
@@ -18,7 +18,7 @@ pub enum CompressedReader {
     #[cfg(feature = "compress")]
     Deflate(DeflateDecoder<BodyReader>),
     #[cfg(feature = "compress")]
-    Gzip(Box<GzDecoder<BodyReader>>),
+    Gzip(GzDecoder<BodyReader>),
```
sbstp: Doesn't the GzDecoder from bufread also need to be boxed?

adamreichold: Avoiding the additional layer of `BufReader` internally seems to make the size difference less pronounced and hence Clippy happy.

src/parsing/chunked_reader.rs
adamreichold: Pushed a fixed version factoring out a `buffers::read_line_ending` helper.

src/parsing/compressed_reader.rs
adamreichold: (`flate2::read::GzDecoder<BodyReader>` is 328 bytes large whereas `flate2::bufread::GzDecoder<BodyReader>` is 296 bytes large.)

src/parsing/chunked_reader.rs
adamreichold: I added a commit that limits the chunk size to `u16` which implies a limit of 64 KB for the chunk buffer.

sbstp: So if a chunk larger than 64 KiB is sent we get an error? I'm not sure if I like that, I'd prefer just splitting the chunk in multiple calls to fill_buf.

adamreichold: > I'd prefer just splitting the chunk in multiple calls to fill_buf.

I can certainly implement that but it is basically the same as implementing any arbitrary fixed amount of buffering which will complicate the implementation as it needs to keep state for handling partially read chunks. As said above, I am wary of the amount of ceremony if it does not buy us anything until e.g. `read_chunk_size` and friends are fixed.

sbstp:
I started thinking about how connection pooling will be implemented, I think the Response will put back a `BufReader<BaseStream>` into the pool when dropped, keyed by the hostname and port. The `BufReader` might contain data from the stream that hasn't been read yet so they cannot be separated. We have to make sure that the `BodyReader` never reads more data than needed, if it does it needs to be put back in the stream somehow.

adamreichold:
> We have to make sure that the BodyReader never reads more data than needed, if it does it needs to be put back in the stream somehow.

I don't anything changed in this regard, i.e. we buffer at most a full chunk and we do not read beyond the last chunk. Of course, one has to throw the connection away if there was an error as we definitely could have consumed more than we wanted, e.g. `buffers::read_line` but this is not affected by this change either.

adamreichold:
(Note that I would suggest to publish the Happy Eyes implementation first and let give this a try before doing another fundamental change like connection pooling.)

sbstp:
Agreed. Should I bump to `0.8` or `0.7.1`? Happy Eyeballs is technically not a breaking change but it's still a big one.

src/parsing/chunked_reader.rs
adamreichold: Pushed a variant that reads at most a single chunk and at most `MAX_BUFFER_LEN` bytes into the internal buffer. I set `MAX_BUFFER_LEN` to 64 kB but tested it using 3 B.

adamreichold:
> Agreed. Should I bump to `0.8` or `0.7.1`? Happy Eyeballs is technically not a breaking change but it's still a big one.

I think bumping this to 0.8 is warranted.

src/parsing/chunked_reader.rs
```diff
@@ -34,95 +33,65 @@ where
...
-            buf.len()
-        );
+            self.buffer.resize(cmp::min(self.remaining, MAX_BUFFER_LEN), 0);
+            self.inner.read_exact(&mut self.buffer)?;
+            self.consumed = 0;
+            self.remaining -= self.buffer.len();
 
-        let n = self.inner.read(&mut buf[..count])?;
-        self.read += n as u64;
+            if self.remaining == 0 && !buffers::read_line_ending(&mut self.inner)? {
```
sbstp: I'm not sure I understand this if, in which will `read_line_ending` return false?

sbstp: Ah if we've read the whole chunk and the next bytes aren't `\r` or `\r\n` it's an error, got it.

sbstp:
I will run this branch through the crawler tomorrow to check for regressions, just to be safe. I'll also push the crawler branch.

sbstp:
No issues with the crawler.

PR #28, opened 2019-12-30
Improve precision of Happy Eyeballs log messages

adamreichold:
Also factors out some common code, avoids the `.copied()` calls and simplifies the receive loops.

### DIFF
--- a/src/happy.rs
+++ b/src/happy.rs
@@ -1,8 +1,7 @@
 use std::io;
 use std::iter::{self, FusedIterator};
-use std::net::TcpStream;
-use std::net::ToSocketAddrs;
-use std::sync::mpsc::{channel, RecvTimeoutError};
+use std::net::{TcpStream, ToSocketAddrs};
+use std::sync::mpsc::channel;
 use std::thread;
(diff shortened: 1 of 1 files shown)


sbstp:
That double loop & match code was bugging me, this is clean, I like it.

src/happy.rs
```diff
@@ -26,73 +25,64 @@ where
...
     }
 
     // We must drop this handle to the sender in order to properly disconnect the channel
     // when all the threads are finished.
     drop(tx);
 
-    // This loop waits for replies from the background threads. It will automatically timeout when the background
-    // thread's connection attempt timeouts and the senders are dropped.
+    // This loop waits for replies from the background threads. It will automatically timeout
+    // when the background thread's connection attempt timeouts and the senders are dropped.
```
sbstp: Can you fix the typo I made earlier? `when the background threads' connection attempts timeout`

adamreichold: Done

sbstp: Thanks, merging.

PR #29, opened 2019-12-30
Add crawler to test with a wide variety of websites and servers

sbstp:


crawler/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "crawler"
+version = "0.1.0"
+authors = ["Simon Bernier St-Pierre <git@sbstp.ca>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
adamreichold: This looks like a left over?

adamreichold:
Do we want to make this an example so it at least it compiled with the tests? Or maybe create workspace? I fear that it might bitrot when it just lies around in the repository.

crawler/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+    let mut listing: Listing = create_req("https://www.reddit.com/r/todayilearned/new.json")
+        .send()?
+        .json()?;
+
+    while let Some(after) = &listing.data.after {
+        if listing.data.children.is_empty() {
+            break;
+        }
+
+        process_items(&listing);
```
adamreichold: Shouldn't we always process the items even if `listing.data.after` is `None`, e.g. for the last listing? So something like
```
let mut listing = create_req(...);

loop {
   process_items(&listing);

  if let Some(after) = listing.data.after {
    listing = create_req(...);
  } else {
    break;
  }
}
```

crawler/src/main.rs
```diff
@@ -0,0 +1,72 @@
+use colored::*;
+use serde::{Deserialize, Serialize};
+
+type Listing = ProxyData<Posts>;
+
+#[derive(Debug, Deserialize, Serialize)]
+struct ProxyData<T> {
```
adamreichold: Why is this generic instead of just containing `pub data: Posts`?

crawler/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+        .header(
+            "User-Agent",
+            "Mozilla/5.0 (Windows NT 6.3; Win64; x64; rv:71.0) Gecko/20100101 Firefox/71.0",
+        )
+        .header("Accept", "*/*")
+}
+
+fn process_items(listing: &Listing) {
+    for item in &listing.data.children {
+        if !item.data.is_self {
```
adamreichold: `if item.data.is_self { continue; }` could reduce rightward drift here.

crawler/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+            match create_req(&item.data.url).send() {
+                Ok(resp) => {
```
adamreichold: Maybe a guard like `Ok(resp) if resp.is_success() => println!(...)` would make this more uniform?

crawler/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+struct Posts {
+    pub children: Vec<ProxyData<Post>>,
+    pub after: Option<String>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+struct Post {
+    pub is_self: bool,
+    pub url: String,
+    pub subreddit: String,
```
adamreichold: Why parse `subreddit` (and `ProxyData::kind`) if it isn't used?

crawler/Cargo.toml
sbstp: This is created by `cargo new`

crawler/src/main.rs
sbstp: I use it at multiple places, in the Listing but also for the children.

crawler/src/main.rs
sbstp: No reason

crawler/src/main.rs
sbstp: Yes you're correct.

crawler/Cargo.toml
adamreichold: But should we leave it here? I usually remove it first thing after creating a new crate.

crawler/src/main.rs
adamreichold: Ah, yes, I see it now.

Issue #32, opened 2020-01-05
Response.text panics when using charsets feature


tdryer:
When the `charsets` feature is enabled, `Response.text` panics in this program:

```
fn main() -> Result<(), attohttpc::Error> {
    let resp = attohttpc::get("https://rust-lang.org/").send()?;
    println!("{}", resp.text()?); // panic here
    Ok(())
}
```

The panic does not occur if the `charsets` feature is disabled, or if `Response.text_utf8` is used instead.

Version:
```
attohttpc = {"version" = "0.8.0", "features" = ["charsets"]}
```

Backtrace:
```
thread 'main' panicked at 'slice index starts at 12288 but ends at 8192', src/libcore/slice/mod.rs:2670:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1028
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1412
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:188
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:205
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:464
  11: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:373
  12: rust_begin_unwind
             at src/libstd/panicking.rs:302
  13: core::panicking::panic_fmt
             at src/libcore/panicking.rs:139
  14: core::slice::slice_index_order_fail
             at src/libcore/slice/mod.rs:2670
[...]

PR #33, opened 2020-01-05
WIP: Fix for TextReader bug #32

sbstp:
Fix for #32

### DIFF
--- a/src/parsing/text_reader.rs
+++ b/src/parsing/text_reader.rs
@@ -50,24 +50,30 @@ impl<R> Read for TextReader<R>
 where
     R: BufRead,
 {
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+    fn read(&mut self, mut buf: &mut [u8]) -> io::Result<usize> {
         if self.eof {
             return Ok(0);
         }
+
+        dbg!(buf.len());
 
         let mut total_written = 0;
 
         loop {
             let src = self.inner.fill_buf()?;
+            dbg!(src.len());
+            dbg!(buf.len());
+
             if src.is_empty() {
                 // inner has reached EOF, write last to the buffer.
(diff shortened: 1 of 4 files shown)


adamreichold:
For reviewability, I think it would be nice to have the bug fix and the rewrite in separate MR.

src/parsing/text_reader.rs
```diff
@@ -28,82 +30,81 @@ where
...
+// }
 
 impl<R> Read for TextReader<R>
 where
     R: BufRead,
 {
-    fn read(&mut self, mut buf: &mut [u8]) -> io::Result<usize> {
-        if self.eof {
+    fn read(&mut self, mut dst: &mut [u8]) -> io::Result<usize> {
+        // The `Read` trait is not directly compatible with the `encoding_rs` crate. There is no
```
adamreichold: But is that really our problem? `Read::read` never guarantees to write the full buffer anyway and passing in a buffer so small that it cannot contain a single code point would then yield the unhelpful but correct result of reading zero bytes.

If we do go for the internal buffer here, then I'd say that it make little sense to implement `io::Read` and we should go for `io::BufRead` instead (with `io::Read` using the trivial implementation as `ChunkedReader` does).

adamreichold: We could also consider just replacing our dependency on `encoding_rs` by [`encoding_rs_io`](https://docs.rs/encoding_rs_io) or if we want to stick to `Read` at least use the same trick as they do with their [`TinyTranscoder`](https://docs.rs/crate/encoding_rs_io/0.1.6/source/src/util.rs), i.e. use a small fixed-size overflow buffer.

adamreichold: Created #34 to show case the `encoding_rs_io` route.

Issue #35, opened 2020-01-07
Inability to configure full request timeout leads to DoS / resource leak


Shnatsel:
attohttpc currently does not allow specifying a timeout for the entire request (i.e. until the request body is finished transferring), which means an established connection will hang indefinitely if the remote host doesn't stops replying. This leads to a resource leak, which may cause denial of service. This post explains the problem in detail: https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779

While it's not necessary to set such timeout by default (most other HTTP clients don't), it must be possible to configure it for attohttpc to be usable for e.g. building reliable services that query a remiote API over HTTP, or for defending against DoS attacks where the remote host sends 1 byte per second, keeping the connection open indefinitely.

Shnatsel:
I have updated the description to clarify the attack vectors and a better approach to mitigating them.

adamreichold:
A read timeout should be easy to add, but a full request timeout is a complicated thing using a blocking design, i.e. one needs to spawn a separate thread that forcibly closes the underlying TCP stream so that all blocking calls return when the timeout expires.

But also not that the DoS problem is mainly one for the server, i.e. a client will probably not have a problem keeping single TCP connection alive, a server which does that on behalf of a lot of clients certainly can get into trouble. But honestly, I don't think attohttpc is a very reasonable choice for such a situation and one would probably prefer a HTTP client that is integrated with the I/O framework of the server itself which will make it reasonably complex to solve things like full request timeouts using e.g. Futures. Or to quote the documentation:
> The intended use is for projects that have HTTP needs where performance is not critical or when HTTP is not the main purpose of the application.

Shnatsel:
I think there is a way to fit it into a blocking design: you can simply configure the read timeout to `min(read_timeout, time_until_request_deadline)`

Shnatsel:
The drawback is that this timeout needs to be reconfigured once in a while, but if it's done e.g. once every 250 milliseconds it's not going to interfere with performance and still provide adequate precision.

adamreichold:
> I think there is a way to fit it into a blocking design: you can simply configure the read timeout to min(read_timeout, time_until_request_deadline)

While such a scheme can be made to work in principle, it is quite brittle as it needs to be threaded through the layers of blocking reads hence complicated code significantly which is why I think that spawning a separate thread would more in line with this crate's goals for simplicity but actually I think the problem is probably no to relevant for this crate's intended use case.

[...]

PR #37, opened 2020-01-07
Extend RequestBuilder to support specifying timeouts

adamreichold:
We expose the connect and read timeouts of the underlying TCP connection and provide reasonable defaults.

Related to #35 but not closing it as it adds a read timeout, a not a request timeout.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -6,6 +6,7 @@ use std::convert::From;
 use std::convert::TryInto;
 use std::io::{prelude::*, BufWriter};
 use std::str;
+use std::time::Duration;
 
 #[cfg(feature = "compress")]
 use http::header::ACCEPT_ENCODING;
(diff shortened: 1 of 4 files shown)


adamreichold:
I think we should only merge one of this or #39, but not both.

src/request.rs
```diff
@@ -108,6 +111,8 @@ impl RequestBuilder {
             body: [],
             max_redirections: 5,
             follow_redirects: true,
+            connect_timeout: Duration::from_secs(30),
```
adamreichold: @sbstp I did not pull this out into a constant to stay consistent with e.g. `max_redirections`. If we want to pull out all of them, I think a separate PR would be nicer to keep this as simple as possible.

sbstp: Yeah that's fine.

sbstp:
Thanks a lot! I'm gonna look into adding the insecure support and I'll publish this with it.

Issue #38, opened 2020-01-08
Allow disabling certificate verification


tscs37:
For development purposes it can be useful to ignore certificate requirements (ie, expired certificates, mismatching hostnames, etc.)

It would be useful if the RequestBuilder had an option to disable Certificate Verification, something like `danger_allow_invalid_certificates()`.

sbstp:
Yes this is a something I wanted to do. I think it can be easily done by adding a parameter to the RequestBuilder i.e. `attohttpc::get("https://localhost:8080").allow_insecure(true)`, something like this.

sbstp:
This is out in 0.10.0

PR #40, opened 2020-01-09
Add RequestBuilder methods to accept invalid TLS

sbstp:


sbstp:
#38

sbstp:
I did some experiments with badssl.com, and I think that `accept_invalid_certs` might be a superset of `accept_invalid_hostnames`, so the third method that sets both would not be required. I'll look further into this.

src/request.rs
```diff
@@ -356,6 +368,50 @@ impl<B> RequestBuilder<B> {
...
+        self.accept_invalid_certs = accept_invalid_certs;
+        self
+    }
+
+    /// Sets if this `Request` will accept an invalid hostname in a TLS certificate.
+    ///
+    /// The default value is `false`.
+    ///
+    /// # Danger
+    /// Use this setting with care. This will accept **any** TLS certificate valid or not.
```
adamreichold: This wording might be changed as this will not accept invalid certifications, just ignore if they do not match the given hostname? Or do you feel one should tone this down in any way due to the security implications?

src/request.rs
```diff
@@ -356,6 +368,50 @@ impl<B> RequestBuilder<B> {
...
+    /// Sets if this `Request` will accept invalid TLS certificates.
+    ///
+    /// The default value is `false`.
+    ///
+    /// # Danger
+    /// Use this setting with care. This will accept **any** TLS certificate valid or not.
+    /// If you are using self signed certificates, it is much safer to add their root CA
+    /// to the list of trusted root CAs by your system.
+    #[cfg(feature = "tls")]
+    pub fn accept_invalid_certs(mut self, accept_invalid_certs: bool) -> Self {
```
adamreichold: I know it looks somewhat ugly, the as the original reporter suggested, one could name this something like `danger_accept_invalid_certs` so it immediately stands out during code review?

src/streams.rs
```diff
@@ -11,6 +11,16 @@ use url::Url;
 use crate::happy;
 use crate::{ErrorKind, Result};
 
+pub struct ConnectInfo<'u> {
```
adamreichold: This is nice! :+1:

tests/test_invalid_certs.rs
```diff
@@ -0,0 +1,58 @@
+#[cfg(feature = "tls")]
```
adamreichold: `tests` in `Cargo.toml` do not support `required-features`?

src/request.rs
sbstp: It should be fixed for the host name. The certificate is still valid but the host name isn't checked.

src/request.rs
sbstp: I personally find the `danger_` a bit patronizing but it seems to be the standard that's been set in the Rust community.

tests/test_invalid_certs.rs
sbstp: Yes it does, for some reason I assumed it didn't.

adamreichold:
@sbstp I was thinking about this again and maybe we want to avoid adding nooks and crannies for all possible network scenarios to attohttpc's API, maybe we just need a public trait like
```
pub trait Streams {
   type Tcp: Read + Write;
   type Tls: Read + Write;

   fn connect_tcp(&mut self, host: &str, port: u16) -> io::Result<Self::Tcp>;
   fn connec_tls(&mut self, host: &str, port: u16) -> io::Result<Self::Tls>;
}
```
provide a default implementation using `happy::connect` and `native_tls` but expose a method like
```
pub struct PreparedRequest {
  pub fn send_with_streams(&mut self, streams: impl Streams) -> Result<Response> { ... }
}
```

This way we can remove the `connect_timeout` and `read_timeout` API as they can be replaced by this. Same goes for this certificate validation handling. It would even allow to use attohttpc with alternative TLS implementations like rustls and we can use this for testing to replace the `BaseStream::Mock` variant.

It would also allow third parties to avoid Happy Eyeballs if they do not want for whatever reason or to add request deadlines in any of the ways we discussed so far without making attohttpc itself much more complex. We would just need to provide documentation and example as guidance on how to do that using this trait.

adamreichold:
#41 is another thing that could be resolved using the API.

sbstp:
What I had in mind for this library was more of a turnkey experience, similar to Python's requests. I think there's a finite and reasonable set of features that people might expect to see in a library like this.

I'm not against offering lower level hooks that enable people to customize the library even more, but for now, I don't want to turn people away and tell them to just do it themselves.

For instance to disable tls verification in requests you just set the `verify` param to `False`.

PR #46, opened 2020-01-17
Only try to redirect for whitelisted status codes

### LINKED ISSUE #30: Client should not redirect on a 300 code
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/300 This multiple choice status code cannot have an automatic follow-up.

### LINKED ISSUE #45: Error returned when parsing 304 Not Modified response
When parsing a 304 Not Modified response, attohttpc returns the error `InvalidResponse(LocationHeader)`.

Version:
```
attohttpc = {"version" = "0.10.1"}
```

Code:
```rust
fn main() -> Result<(), attohttpc::Error> {
    let resp = attohttpc::get("http://httpbin.org/status/304").send()?;
    println!("{}", resp.text()?);
    Ok(())
}
```

Output:
```rust
Error: Error(InvalidResponse(LocationHeader))
```

tdryer:
The current version of attohttpc tries to redirect for any 3xx status code. However, not all such status codes indicate an automatically processable redirect. Instead, only try to redirect for whitelisted status codes.

Fixes #30. Fixes #45.

### DIFF
--- a/tests/test_redirection.rs
+++ b/tests/test_redirection.rs
@@ -1,6 +1,4 @@
-use std::net::TcpStream;
 use std::thread;
-use std::time::{Duration, Instant};
 
 use attohttpc::ErrorKind;
 use lazy_static::lazy_static;
@@ -15,18 +13,6 @@ lazy_static! {
         )).unwrap();
         let port = server.server_addr().port();
         thread::spawn(|| { server.run(); });
-
-        let start = Instant::now();
-        let timeout = Duration::from_secs(10);
-
-        // Wait until server is ready. 10s timeout in case of error creating server.
-        while TcpStream::connect(("localhost", port)).is_err() {
-            if start.elapsed() > timeout {
-                panic!("time out in server creation");
-            }
-            thread::sleep(Duration::from_millis(100));
-        }
-
         port
     };
 }
--- a/tests/test_redirection.rs
+++ b/tests/test_redirection.rs
@@ -4,12 +4,16 @@ use std::time::{Duration, Instant};
 
 use attohttpc::ErrorKind;
 use lazy_static::lazy_static;
-use rouille::Response;
+use rouille::{router, Response};
 
 lazy_static! {
     static ref STARTED: bool = {
         thread::spawn(move || {
-            rouille::start_server("localhost:55123", move |_| Response::redirect_301("/"));
+            rouille::start_server("localhost:55123", move |request| router!(request,
+                (GET) ["/301"] => Response::redirect_301("/301"),
+                (GET) ["/304"] => Response::text("").with_status_code(304),
+                _ => Response::empty_404()
+            ))
         });
 
         let start = Instant::now();
@@ -31,7 +35,7 @@ lazy_static! {
 fn test_redirection_default() {
     let _ = *STARTED;
 
-    match attohttpc::get("http://localhost:55123/").send() {
+    match attohttpc::get("http://localhost:55123/301").send() {
         Err(err) => match err.kind() {
             ErrorKind::TooManyRedirections => (),
             _ => panic!(),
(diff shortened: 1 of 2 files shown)


tdryer:
I copied the [list of redirecting status codes from Requests](https://github.com/psf/requests/blob/428f7a275914f60a8f1e76a7d69516d617433d30/requests/models.py#L48).

tests/test_redirection.rs
```diff
@@ -4,12 +4,16 @@ use std::time::{Duration, Instant};
...
 use attohttpc::ErrorKind;
 use lazy_static::lazy_static;
-use rouille::Response;
+use rouille::{router, Response};
 
 lazy_static! {
     static ref STARTED: bool = {
         thread::spawn(move || {
-            rouille::start_server("localhost:55123", move |_| Response::redirect_301("/"));
+            rouille::start_server("localhost:55123", move |request| router!(request,
```
adamreichold: I wonder if using `rouille` is really necessary here. Using a fixed response should be sufficient and binding on a free random instead of fixed high port seems advisable, c.f. `test_timeout.rs`. This is the only place we are using this and lLoosing `rouille` would lighten the development dependencies significantly which would be nice IMHO.

sbstp:
I had started working on this slowly in a branch and I was trying to handle 303 see other but honestly I prefer the way you did this. See other is kind of a pain in the ass to implement, we should just let users handle it in a way that works for them.

tests/test_redirection.rs
sbstp: I personally don't mind depending on rouille for development only. It's important for me to have end to end tests. I also think it's easier to write them this way.

I agree on the port thing though, it would be preferable to use a random unallocated port.

tdryer:
Updated `test_redirection.rs` to get an unused port like `test_timeout.rs` does.

tests/test_redirection.rs
```diff
@@ -1,37 +1,43 @@
...
 
 use attohttpc::ErrorKind;
 use lazy_static::lazy_static;
-use rouille::Response;
+use rouille::{router, Response};
 
 lazy_static! {
-    static ref STARTED: bool = {
+    static ref STARTED: u16 = {
+        let port = TcpListener::bind("localhost:0").unwrap().local_addr().unwrap().port();
```
adamreichold: Isn't this racy without explicitly setting `SO_REUSE` port?

tests/test_redirection.rs
adamreichold: > It's important for me to have end to end tests. 

I totally agree. But we can have those and fix the port thing by dropping `rouille` and add single call to `stream.write_all(b"some hard-coded HTTP response").unwrap();` into the `test_timeout.rs` example.

adamreichold: We could also avoid the `// Wait until server is ready. 10s timeout in case of error creating server.` part as we could bind the socket on the main thread only put the accept loop onto a background thread.

adamreichold: Create an example commit on how we could do that: https://github.com/adamreichold/attohttpc/commit/580abf0d66c17b226ec8214b4b0a91cd64dd1e1b

tests/test_redirection.rs
sbstp: Is there anyway to provide rouille with the TcpListener to use? I think this might be racy yes.

adamreichold: > Is there anyway to provide rouille with the TcpListener to use? I think this might be racy yes.

The default recipe I know of is to bind using `SO_REUSEPORT` and keep the initial listener alive while starting the second one but dropping it before establishing connections. But honestly speaking, I think adamreichold@580abf0 is much simpler.

adamreichold: Last commit did resolve the race. @tdryer I think it should also allow us to remove the waiting loop since we only spawn the thread after the socket is bound?

tdryer: Yep, the waiting loop is unnecessary now. I'll remove it.

sbstp:
Thanks to both of you for this!

PR #47, opened 2020-01-19
Add support for a full request timeout

### LINKED ISSUE #35: Inability to configure full request timeout leads to DoS / resource leak

adamreichold:
By racing another thread which shuts down the socket when the timeout is elapsed.

Closes #35

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,9 +1,7 @@
 #![allow(dead_code)]
 #![allow(clippy::write_with_newline)]
-use std::borrow::Borrow;
-use std::borrow::Cow;
-use std::convert::From;
-use std::convert::TryInto;
+use std::borrow::{Borrow, Cow};
+use std::convert::{From, TryInto};
 use std::io::{prelude::*, BufWriter};
 use std::str;
 use std::time::Duration;
@@ -72,6 +70,7 @@ pub struct RequestBuilder<B = [u8; 0]> {
     follow_redirects: bool,
     connect_timeout: Duration,
     read_timeout: Duration,
+    timeout: Option<Duration>,
     #[cfg(feature = "charsets")]
     pub(crate) default_charset: Option<Charset>,
     #[cfg(feature = "compress")]
@@ -117,6 +116,7 @@ impl RequestBuilder {
             follow_redirects: true,
             connect_timeout: Duration::from_secs(30),
             read_timeout: Duration::from_secs(30),
+            timeout: None,
             #[cfg(feature = "charsets")]
             default_charset: None,
             #[cfg(feature = "compress")]
@@ -256,6 +256,7 @@ impl<B> RequestBuilder<B> {
             follow_redirects: self.follow_redirects,
             connect_timeout: self.connect_timeout,
             read_timeout: self.read_timeout,
+            timeout: self.timeout,
             #[cfg(feature = "charsets")]
             default_charset: self.default_charset,
             #[cfg(feature = "compress")]
@@ -349,6 +350,14 @@ impl<B> RequestBuilder<B> {
         self
     }
 
+    /// Sets a timeout for the whole request.
+    ///
+    /// Applies after a TCP connection is established. Defaults to no timeout.
+    pub fn timeout(mut self, duration: Duration) -> Self {
+        self.timeout = Some(duration);
+        self
+    }
+
     /// Set the default charset to use while parsing the response of this `Request`.
     ///
     /// If the response does not say which charset it uses, this charset will be used to decode the request.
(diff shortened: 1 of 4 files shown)


sbstp:
Hey! Thanks for this. It's a clean way of doing it. Do you think that we could use like an AtomicBool or something like that to signal the BaseStream that the UnexpectedEOF was caused by the request timeout? Then we can return a timeout error instead of UnexpectedEOF.

adamreichold:
> Then we can return a timeout error instead of UnexpectedEOF.

Yes, but we would have keep that flag around (`Arc<AtomicBool>` to ensure a sufficiently long lifetime, adding a field to `BaseStream`) and we would have to check it after every failed read.

Personally, I think the code complexity this adds is not worth the change to the more relevant error code (`io::ErrorKind` is already rather vague as it is, e.g. `WouldBlock` and `TimedOut` depending on platform), especially since simplicity is the whole point of this approach.

adamreichold:
> we would have to check it after every failed read.

Actually worse, we have to check every read as this yields a "clean" read of length zero.

sbstp:
I don't really mind carrying an Arc<AtomicBool> around. I think we could map on the error returned by `parse_response` and add a method to the BaseStream to check if the thread timedout.

```rust
let mut stream = BaseStream::connect(&info)?;
self.write_request(&mut stream, &url)?;
let resp = match parse_response(stream, self).map_err(|err| {
    if let ErrorKind::Io(io_err) = err.kind() {
        if let io::ErrorKind::UnexpectedEof = io_err.kind() {
            if stream.timeout_reached() {
                return ErrorKind::Timeout;
            }
        }
    }
    err
});
```

We could also use a channel and recv_timeout to end the thread sooner if the request finishes quickly. On drop BaseStream could send something to wake up the timeout thread and tell it to end.

sbstp:
You don't actually need to send anything in the channel. On drop it will get disconnected.
```rust
match rx.recv_timeout(timeout) {
  Err(RecvTimeoutError::Timeout) => stream.shutdown(Shutdown::Both),
  _ => { // end thread }
}
```

sbstp:
Forget what I said it doesn't make sense. You can still read the body after the parse_response call, you could timeout while reading the body.

adamreichold:
> We could also use a channel and recv_timeout to end the thread sooner if the request finishes quickly. 

I have implemented this and pushed it as separate commit. It is you prerogative to make this choice, but personally I still think the complexity cost is completely unproportional for such a fringe feature.

sbstp:
Let's take some time to think about it. Even if we decide to remove the error check in the BaseStream, I think we should keep the channel to terminate the thread early when the Response is dropped quickly. If for some reason someone puts a really large timeout, we start leaking threads.

adamreichold:
> If for some reason someone puts a really large timeout, we start leaking threads.

Using a single thread that just sleeps until the next deadline might be a more resource efficient solution but not we are talking about inventing a timer wheel...

sbstp:
Looks like the pipeline failed. Race condition? Spurious? Genuine failure?

adamreichold:
> Looks like the pipeline failed. Race condition? Spurious? Genuine failure?

Spurious but also genuine: I it necessary to explicitly drop the receiver as otherwise we can still receive another item after having called `shutdown` so that sending will not fail and we will not consider this a timeout. But then again, the failure mode is replacing one I/O error kind by another...

sbstp:
I'm ok with removing the code that converts the error from UnextectedEof to a timeout, but I'd like to keep the channel that closes the thread as soon as the response is dropped.

adamreichold:
> I'm ok with removing the code that converts the error from UnextectedEof to a timeout, but I'd like to keep the channel that closes the thread as soon as the response is dropped.

Done. (The changes in the redirection test are due to `cargo fmt`. Maybe we should include checking the formatting in the CI?)

adamreichold:
Sorry for the CI noise, I messed up the test during rebase.

sbstp:
Thanks! I'll release this & the redirect fix right away.

PR #50, opened 2020-01-27
Make sure to timeouts do not look like normal EOF.

### LINKED ISSUE #49: Timeout silently truncates response body
Specifying `timeout` for a request appears to silently truncate the response body if it being received when the the timeout expires.

Example:

`http://httpbin.org/drip` will return 10 bytes over 2 seconds:
```sh
$ time curl "http://httpbin.org/drip"; echo
**********
real    0m2.005s
```

With a 1 second timeout, curl returns an error:
```sh
$ time curl -m 1 "http://httpbin.org/drip"; echo
curl: (28) Operation timed out after 1000 milliseconds with 4 out of 10 bytes received
****
real    0m1.008s
```

This is an attohttpc program that I expected do the same thing:

```rust
use std::time::Duration;

use attohttpc::Result;

fn main() -> Result {
    let resp = attohttpc::get("http://httpbin.org/drip")
        .timeout(Duration::from_secs(1))
        .send()?;
    println!("Status: {:?}", resp.status());
    println!("Body:\n{}", resp.text()?);
    Ok(())
}
```

But instead of returning an error, the result is successful, and `text` is truncated:

```sh
$ time cargo run --example drip
Status: 200
Body:
*****

real    0m1.178s
```

adamreichold:
Closes #49

### DIFF
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -68,6 +68,7 @@ impl BaseStream {
                 let (tx, rx) = mpsc::channel();
                 thread::spawn(move || {
                     if let Err(mpsc::RecvTimeoutError::Timeout) = rx.recv_timeout(timeout) {
+                        drop(rx);
                         let _ = stream.shutdown(Shutdown::Both);
                     }
                 });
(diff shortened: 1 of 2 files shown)


sbstp:
Isn't this what we decided to abandon in your last merge request?

adamreichold:
> Isn't this what we decided to abandon in your last merge request?

Indeed. But now we got a bug report about this being unexpected behavior. So I bit my teeth and tried to implement something reasonable.

The alternative seems to be to tell the user to check his/her content length which does not sound very convincing. (Or to implement those checks ourselves everywhere but even this would leave a discrepancy between the `io::Read`-based interface and high-level ones like `.text()`.)

sbstp:
I care more about giving a under friendly library than a fast library, so I'm going to merge this. I don't think this will affect performance much anyway, we're doing IO after all.

sbstp:
I'll release a patch with this fix.

adamreichold:
> I care more about giving a under friendly library than a fast library, so I'm going to merge this. I don't think this will affect performance much anyway, we're doing IO after all.

I agree. Also I would like to add that my personal reservations against this code where never about performance, it is complexity I am concerned about, i.e. mostly readability and learnability of the code base.

Issue #41, opened 2020-01-10
Allow to specify certificates for client authentication


silvioprog:
Hi.

Do you have plans to allow to specify certificates (`.pem`, `.pfx`, `.p12` etc.) for client authentication?

I did a small change in file `streams.rs` ([line 51](https://github.com/sbstp/attohttpc/blob/73d650b35e615847f61c26a30114eb70429e1f97/src/streams.rs#L51)) just to test a connection which requires a PKCS12 certificate:

```rust
...
    #[cfg(feature = "tls")]
    fn connect_tls(
        host: &str,
        port: u16,
        connect_timeout: Duration,
        read_timeout: Duration,
    ) -> Result<TlsStream<TcpStream>> {
        use native_tls::Identity;
        let mut builder = TlsConnector::builder();
        let buf = std::fs::read("/home/user/certificate.pfx")?;
        let pkcs12 = Identity::from_pkcs12(&buf, "123456").unwrap();
        builder.identity(pkcs12);
        builder.danger_accept_invalid_certs(true);
        let connector = builder.build()?;
        let stream = BaseStream::connect_tcp(host, port, connect_timeout, read_timeout)?;
        let tls_stream = match connector.connect(host, stream) {
            Ok(stream) => stream,
            Err(HandshakeError::Failure(err)) => return Err(err.into()),
            Err(HandshakeError::WouldBlock(_)) => panic!("socket configured in non-blocking mode"),
        };
        Ok(tls_stream)
    }
...
```

and it worked fine.

(The `danger_accept_invalid_certs(true)` is related to #38).

adamreichold:
This is now possible when using the `rustls` backend as we expose the full `ClientConfig`. For `native-tls`, we either need to add API or also expose `TlsConnector` directly.

PR #51, opened 2020-01-31
Add optional support for rustls as an alternative to native-tls

### LINKED ISSUE #48: Offer rustls as an alternative to native-tls
As suggested in the [smoke testing blog post](https://medium.com/@shnatsel/smoke-testing-rust-http-clients-b8f2ee5db4e6) we could offer to use the `rustls` TLS backend instead of `native-tls` with a feature flag.

adamreichold:
Closes #48

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -131,3 +131,22 @@ where
 {
     RequestBuilder::new(Method::TRACE, base_url)
 }
+
+mod skip_debug {
+    use std::fmt;
+
+    #[derive(Clone)]
+    pub struct SkipDebug<T>(pub T);
+
+    impl<T> fmt::Debug for SkipDebug<T> {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(f, "...")
+        }
+    }
+
+    impl<T> From<T> for SkipDebug<T> {
+        fn from(val: T) -> SkipDebug<T> {
+            SkipDebug(val)
+        }
+    }
+}
--- a/src/request.rs
+++ b/src/request.rs
@@ -4,6 +4,8 @@ use std::borrow::{Borrow, Cow};
 use std::convert::{From, TryInto};
 use std::io::{prelude::*, BufWriter};
 use std::str;
+#[cfg(feature = "tls-rustls")]
+use std::sync::Arc;
 use std::time::Duration;
 
 #[cfg(feature = "compress")]
@@ -12,12 +14,16 @@ use http::{
     header::{HeaderValue, IntoHeaderName, ACCEPT, CONNECTION, CONTENT_LENGTH, HOST, USER_AGENT},
     HeaderMap, Method, StatusCode, Version,
 };
+#[cfg(feature = "tls-rustls")]
+use rustls::ClientConfig;
 use url::Url;
 
 #[cfg(feature = "charsets")]
 use crate::charsets::Charset;
 use crate::error::{Error, ErrorKind, InvalidResponseKind, Result};
 use crate::parsing::{parse_response, Response};
+#[cfg(feature = "tls-rustls")]
+use crate::skip_debug::SkipDebug;
 use crate::streams::{BaseStream, ConnectInfo};
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -79,6 +85,8 @@ pub struct RequestBuilder<B = [u8; 0]> {
     accept_invalid_certs: bool,
     #[cfg(feature = "tls")]
     accept_invalid_hostnames: bool,
+    #[cfg(feature = "tls-rustls")]
+    client_config: SkipDebug<Option<Arc<ClientConfig>>>,
 }
 
 impl RequestBuilder {
@@ -125,6 +133,8 @@ impl RequestBuilder {
             accept_invalid_certs: false,
             #[cfg(feature = "tls")]
             accept_invalid_hostnames: false,
+            #[cfg(feature = "tls-rustls")]
+            client_config: None.into(),
         })
     }
 }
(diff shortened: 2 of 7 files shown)


src/lib.rs
```diff
@@ -131,3 +131,22 @@ where
 {
     RequestBuilder::new(Method::TRACE, base_url)
 }
+
+mod skip_debug {
+    use std::fmt;
+
+    #[derive(Clone)]
+    pub struct SkipDebug<T>(pub T);
```
adamreichold: This is somewhat hacky IMHO but I saw few alternatives and those have significant downsides:
* Use a procedural macro like [`educe`](https://crates.io/crates/educe) but that adds an additional dependency and increases build times.
* Manually implement `fmt::Debug` for the relevant types but that seems to imply an unreasonable amount of boiler plate.

src/streams.rs
```diff
@@ -51,9 +68,9 @@ impl BaseStream {
                 BaseStream::connect_tcp(host, port, info).map(|(stream, timeout)| BaseStream::Plain { stream, timeout })
             }
             #[cfg(feature = "tls")]
-            "https" => {
-                BaseStream::connect_tls(host, port, info).map(|(stream, timeout)| BaseStream::Tls { stream, timeout })
-            }
+            "https" => BaseStream::connect_tls(host, port, info),
+            #[cfg(feature = "tls-rustls")]
```
adamreichold: Adding it this way might be surprising but allows one to build a functional instance of this crate with both features enabled which might happen if one depends on this indirectly with the dependencies in between enabling both features. Alternative one could force a compiler error when both features are enabled to prevent this situation from happening.

This will produce a warning about the pattern being unreachable in that case but should function normally preferring the existing `native-tls`-based code path. One could even consider the warning a feature but IIRC, Cargo hides warnings from dependencies by default...

src/request.rs
```diff
@@ -79,6 +85,8 @@ pub struct RequestBuilder<B = [u8; 0]> {
     accept_invalid_certs: bool,
     #[cfg(feature = "tls")]
     accept_invalid_hostnames: bool,
+    #[cfg(feature = "tls-rustls")]
+    client_config: SkipDebug<Option<Arc<ClientConfig>>>,
```
adamreichold: This differs slightly in the approach compared to the existing `accept_invalid_*` fields but is strictly more general as it supports e.g. setting up client authentication. I could also reuse the existing fields marking them as `#[cfg(any(feature = "tls", feature  = "tls-rustls"))]` and implementing their functionality in terms of `rustls::ClientConfig` if you prefer.

Another option I see here would be to align this by replacing the `accept_invalid_*` fields by a `tls_connector: Option<TlsConnector>` which would also allow full access to the `native-tls` configuration but would imply an incompatible API change.

adamreichold: Going for `tls_connector: Option<TlsConnector>` would also allow us to close #41 so this does look like a nice choice.

src/streams.rs
```diff
@@ -100,7 +113,33 @@ impl BaseStream {
                 }
             },
         };
-        Ok((stream, timeout))
+        Ok(BaseStream::Tls { stream, timeout })
+    }
+
+    #[cfg(feature = "tls-rustls")]
+    fn connect_rustls(host: &str, port: u16, info: &ConnectInfo) -> Result<BaseStream> {
+        let name = DNSNameRef::try_from_ascii_str(host)?;
```
sbstp: It seems like this crate is still pretty low level. Do you think there's value in abstracting tls functionality in a module that supports multiple backends, but offers an identical/abstract interface to use tls functionality?

adamreichold: Personally, I did not find it significantly worse than using `native-tls` and there is also the `rustls-connector` crate that provides almost exactly the same API and I did use that in the beginning but the results was not that much simpler.

Considering an additional internal abstraction over `rustls` and `native-tls` more specifically: I think that would only make sense if we decide to avoid exposing the `TlsConnector` or `ClientConfig` for arbitrary TLS setup via our API and provide our own configuration API instead.  Abstracting only the interaction with `TlsStream` and `StreamOwned` does not seem worth it IMHO as `BaseStream` basically already provides exactly this.

sbstp:
Hey I experimented a little and I'd like to make the `tls` features mutually exclusive. I came up with this:

```rust
#[cfg(all(feature = "tls", feature = "tls-rustls"))]
compile_error!("cannot activate both the `tls` and `tls-rustls` feature at once");
```

in `src/lib.rs`

adamreichold:
> I'd like to make the tls features mutually exclusive.

As written above, I know the technique but I think it is quite problematic: Consider you create a program that depends on libraries A and B which both depend on attohttpc. A enables the `tls` features and does not give you control over this whereas B enables the `tls-rustls` feature and gives you no control over this. Hence you'll end up with a dependency tree where both features are enabled without being able to do anything about it short of modifying one of your dependencies. 

Considering that Cargo features are meant to be additive and to avoid giving the user a compiler error that cannot be fixed, I would suggest avoiding this. Especially since there is no technical reason to do so: The library will work with both TLS backends builtin and both provide reasonable defaults (and both would need to be setup correctly in the above scenario anyway).

sbstp:
Hey, sorry for the delay. I wanted to slow down the pace a bit because we've been releasing multiple versions per week. I'll merge this soon. I've also been working on a Session type that can apply settings to all requests made from it and eventually support a cookie jar.

src/request.rs
```diff
@@ -407,6 +419,16 @@ impl<B> RequestBuilder<B> {
         self.accept_invalid_hostnames = accept_invalid_hostnames;
         self
     }
+
+    /// Sets the TLS client configuration
+    ///
+    /// Defaults to a configuration using the root certificates
+    /// from the webpki-roots crate.
+    #[cfg(feature = "tls-rustls")]
+    pub fn client_config(mut self, client_config: Arc<ClientConfig>) -> Self {
```
sbstp: Would it be a good idea to take a `impl Into<Arc<ClientConfig>>` to make the boxing to Arc automatic if you don't intend to share the structure?

sbstp:
Thanks Adam. I ran the reddit crawler on this version and it ran smoothly so I'm confident to merge this.

PR #52, opened 2020-03-01
Add Session type which can apply settings to requests created from it

sbstp:


sbstp:
Review: @adamreichold

sbstp:
Going to do a pass on the docs to make sure everything is still coherent.

sbstp:
@adamreichold hey I was looking at the code and it seems like I missed something. Does rustls listen to the danger_accept_invalid_certs flag?

adamreichold:
> hey I was looking at the code and it seems like I missed something. Does rustls listen to the danger_accept_invalid_certs flag?

It does not. It gets all of its configuration via the `ClientConfig` data type which is immutable because it is usually shared process-wide. Also note that those two flags do not exist when only the rustls backend is configured. As said in the original rustls MR, I see two options here: Implement API surface abstracting over native-tls and rustls covering the options we feel necessary to expose or align things by exposing `TlsConnector` directly so that library users can use that to configure things as they wish.

.travis.yml
```diff
@@ -5,3 +5,5 @@ rust:
   - nightly
 script:
   - /bin/sh run-tests.sh
+  - rustup component add clippy
```
adamreichold: I wonder if this could land separately?

src/request/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+    Error: From<V::Error>,
+{
+    let value = value.try_into()?;
+    headers.append(header, value);
+    Ok(())
+}
+
+/// Represents a request that's ready to be sent. You can inspect this object for information about the request.
+#[derive(Debug)]
+pub struct PreparedRequest<B> {
```
adamreichold: If the request builder is now called `RequestBuilder`, this could become `Request`?

sbstp: I didn't change the name of the RequestBuilder, it was already named this way. The PreparedRequest terminology comes from the `requests` library where a request ready to be sent is available in immutable form to inspect before sending.

.travis.yml
sbstp: Yes, I forgot to add cargo clippy --all-features anyway.

adamreichold: We probably want both, a run with `--no-default-features` and a run with `--all-features`?

src/request/mod.rs
adamreichold: Ah, sorry, seems I got confused here.

PR #54, opened 2020-03-11
Do not try to resolve bracketed IP addresses

### LINKED ISSUE #53: Support for IPv6 addresses
While switching my code from IPv4 only to mixed mode as well as IPv6 only mode, I encountered a strange bug in *attohttpc*.

At first, I thought it would be as easy as replacing strings such as *"http://127.0.0.1:3000/fancy_rest"* with *"http://[::1]:3000/fancy_rest"*. The latter works in my browser and in `curl`. Even the `url` crate can parse it [correctly](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=025db104ed5330d728b9c7752c473333). 

Just when I use `attohttpc` I get the following error on Linux:
```
Error(Io(Custom { kind: Other, error: "failed to lookup address information: Name or service not known" }))
```
on other systems it might also be displayed like the following:
```
Error(Io(Custom { kind: Other, error: "failed to lookup address information: Temporary failure in name resolution" })) 
```

Probably trying to interpret *[::1]* as a hostname instead of an `IpAddr`.

Unfortunately, `attohttpc` is not yet available on the rust playground, otherwise, I would have shown a minimal example there.

And totally unrelated: thank you for providing a pretty neat, slim and efficient http client library without async+await :D

adamreichold:
URL can contain IP addresses within brackets like "http://[::1]" which should
not be resolve as a domain name but rather used as-is as an IP address.

This change therefore makes use of Url::host instead of Url::host_str to determine
the address to connect with whereas Url::host_str continues to be used for TLS
authentication as this is always based on domain names.

Closes #53

### DIFF
--- a/src/happy.rs
+++ b/src/happy.rs
@@ -1,9 +1,11 @@
 use std::io;
 use std::iter::{self, FusedIterator};
-use std::net::{TcpStream, ToSocketAddrs};
+use std::net::{TcpStream, IpAddr, ToSocketAddrs};
 use std::sync::mpsc::channel;
 use std::thread;
 use std::time::{Duration, Instant};
+
+use url::Host;
 
 const RACE_DELAY: Duration = Duration::from_millis(200);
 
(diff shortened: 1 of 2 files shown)


adamreichold:
@gh0st42 Could you give this a try?

gh0st42:
Issue seems to be resolved!

Thanks

sbstp:
Hello! Thanks for the report & the fix. I just published this in version `0.12.0`.

PR #57, opened 2020-04-17
Do some maintenance chores

adamreichold:
This includes Rustfmt and Clippy as part of the CI, updates the formatting and updates the optional Rustls-related dependencies.


src/streams.rs
```diff
@@ -55,6 +55,7 @@ impl BaseStream {
 
         debug!("trying to connect to {}:{}", host, port);
 
+        #[allow(unreachable_patterns)]
```
sbstp, line 58: Neat!

PR #58, opened 2020-04-17
Add Response::error_for_status method to simplify checking responses

### LINKED ISSUE #56: Add helper for converting Response into Result based on StatusCode
When `Response` is received it can be checked for success using `is_success()` method in order to decide on further processing. It may be very helpful to provide a method that will convert `Response` into `Result<Response>` based on the the `StatusCode`.

Other HTTP client libraries provide similar facilities and they are very handy.
For example Python `requests` has [`raise_for_status`](https://requests.readthedocs.io/en/master/api/#requests.Response.raise_for_status) and `reqwest` has [`error_for_status`](https://docs.rs/reqwest/0.10.4/reqwest/struct.Response.html#method.error_for_status).

What do you think?

adamreichold:
Closes #56

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -61,6 +61,8 @@ pub enum ErrorKind {
     InvalidResponse(InvalidResponseKind),
     /// Too many redirections
     TooManyRedirections,
+    /// Status code indicates failure
+    StatusCode(http::StatusCode),
     /// JSON decoding/encoding error.
     #[cfg(feature = "json")]
     Json(serde_json::Error),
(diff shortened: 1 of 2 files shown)


sbstp:
Is it worth also adding the `ref` variant? https://docs.rs/reqwest/0.10.4/reqwest/struct.Response.html#method.error_for_status_ref

adamreichold:
> Is it worth also adding the ref variant?

I did wonder about that but came to the conclusion that the by-value version provides more safety as one cannot access a failed response after turning it into a `Result`. Since the by-ref variant could just as well be open coded, I preferred to keep it safe and simple.

That said, I have to real problem with adding that if you would like to have it.

sbstp:
Fine like this. If someone wants the feature we can iterate later.

Issue #55, opened 2020-04-10
Support streaming request bodies


sfackler:
It'd be nice if there were a way to send a request body without needing to buffer the entire thing in memory first.

adamreichold:
@sbstp What do you think of porting zeptohttpc's [`BodyWriter`](https://github.com/adamreichold/zeptohttpc/blob/master/src/body_writer.rs) trait here?

sbstp:
@adamreichold looks like a good idea.

sbstp:
Could we support streaming bodies for non Seek streams also? Just use like the CompressedBody does.

adamreichold:
> Could we support streaming bodies for non Seek streams also? Just use like the CompressedBody does.

AFAIU the problem for pure `Read` bodies is that with redirects, one might need to produce the body multiple times and hence the need to rewind. `CompressedBody` is just a wrapper that relies on the inner body to rewind and `JsonBody` does not consume the actual value that is written.

I see two alternatives here:
* The body is able to indicate whether it can rewind or not and redirects just fail if the body was already consumed but that seems rather brittle to me.
* We provide an implementation based on `F: Fn() -> R where R: Read` in stead of `R: Read`, i.e. a body that can produce a new readable stream every time but that might end up being more complicated than just letting the user reimplement `BodyWriter` directly.

sbstp:
We should look at the HTTP semantics for this. I think we could implement the Expect-Continue mechanism to solve this elegantly.

adamreichold:
> We should look at the HTTP semantics for this. I think we could implement the Expect-Continue mechanism to solve this elegantly.

For a conforming server implementation, yes. More home-grown server implementations may wait for the body before producing the redirect response. But personally, I am not sure how relevant those are and if we should invest any effort into automatic downgrading.

But more generally speaking, do you have a use case in mind where we have a pure `Read` body that can only be consumed once? File contents and JSON values are possible without this. The client part of a proxy implementation would be the only thing I can come up with right now and I am not sure attohttpc would be a good choice for that.

sbstp:
It would be a nice to have, but it's certainly not a blocker for this to land. We can start small and figure out later how to add it if someone asks for it.

adamreichold:
Also the following part of the specification could be read like us need to be able to produce a body in any case.

[...]

PR #59, opened 2020-04-17
Support generic bodies using a dedicated trait

adamreichold:
This adds the `body` module to generically describe rewindable bodies which is the first step towards resolving #55. (We still need to implement a chunked transfer encoding writer to fully support streaming bodies, but that can be done as a follow-up change.)

### DIFF
--- a/src/request/body.rs
+++ b/src/request/body.rs
@@ -1,0 +1,64 @@
+use std::convert::TryInto;
+use std::io::{Result as IoResult, Write};
+
+/// The kinds of request bodies currently supported by this crate.
+#[derive(Debug, Clone, Copy)]
+pub enum BodyKind {
+    /// An empty request body
+    Empty,
+    /// A request body with a known length
+    KnownLength(u64),
+}
+
+/// A generic rewindable request body
+pub trait Body {
+    /// Determine the kind of the request body
+    fn kind(&mut self) -> IoResult<BodyKind>;
+
+    /// Write out the request body into the given writer
+    fn write<W: Write>(&mut self, writer: W) -> IoResult<()>;
+}
+
+/// An empty request body
+#[derive(Debug, Clone, Copy)]
+pub struct Empty;
+
+impl Body for Empty {
+    fn kind(&mut self) -> IoResult<BodyKind> {
+        Ok(BodyKind::Empty)
+    }
+
+    fn write<W: Write>(&mut self, _writer: W) -> IoResult<()> {
+        Ok(())
+    }
+}
+
+/// A request body containing UTF-8-encoded text
+#[derive(Debug, Clone)]
+pub struct Text<B>(pub B);
+
+impl<B: AsRef<str>> Body for Text<B> {
+    fn kind(&mut self) -> IoResult<BodyKind> {
+        let len = self.0.as_ref().len().try_into().unwrap();
+        Ok(BodyKind::KnownLength(len))
+    }
+
+    fn write<W: Write>(&mut self, mut writer: W) -> IoResult<()> {
+        writer.write_all(self.0.as_ref().as_bytes())
+    }
+}
+
+/// A request body containing binary data
+#[derive(Debug, Clone)]
+pub struct Bytes<B>(pub B);
+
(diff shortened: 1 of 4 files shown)


src/request/body.rs
```diff
@@ -0,0 +1,77 @@
...
+#[derive(Debug)]
+pub struct File(pub fs::File);
+
+impl Body for File {
+    fn kind(&mut self) -> IoResult<BodyKind> {
+        self.0.seek(SeekFrom::End(0)).map(BodyKind::KnownLength)
+    }
+
+    fn write<W: Write>(&mut self, mut writer: W) -> IoResult<()> {
+        copy(&mut self.0, &mut writer).map(|_| ())
```
sbstp, line 75: Should you seek to the start of the file before copying?

src/request/body.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+/// A request body containing UTF-8-encoded text
+#[derive(Debug, Clone)]
+pub struct Text<B>(pub B);
+
+impl<B: AsRef<str>> Body for Text<B> {
+    fn kind(&mut self) -> IoResult<BodyKind> {
+        Ok(BodyKind::KnownLength(self.0.as_ref().len().try_into().unwrap()))
```
sbstp, line 43: I'm a bit puzzled by the fact that there no `From<usize> for u64`. `usize` is either 32 or 64 bits and `u64` is always 64 bits... It should be possible.

src/request/body.rs
adamreichold: Indeed. That is an oversight during porting that code.

src/request/body.rs
adamreichold: There is no such implementation which future-proofs Rust for 128-bit point width targets, but the current implementation is basically
```
#[cfg(target_pointer_width = "64")]
try_from_unbounded!(usize, u64, u128);
```
and hence will never fail.  So while this is quite verbose, there is no associated runtime overhead.

src/request/body.rs
adamreichold: Fixed.

sbstp:
Do you intend on eventually (in another PR) adding support for chunked bodies and allowing people to implement the body trait? I think we could avoid allocating in the `json` and `form` case if we did. Also we'd permit people to stream arbitrary stuff into the request.

adamreichold:
> Do you intend on eventually (in another PR) adding support for chunked bodies and allowing people to implement the body trait?

As written in the cover letter 

> (We still need to implement a chunked transfer encoding writer to fully support streaming bodies, but that can be done as a follow-up change.)

So yes, I do plan to add this in another follow-up MR, but as it separates nicely, splitting it up makes it easier to review. User code should also already be able to implement the trait and the only thing left to make this fully work is to make the `RequestBuilder::body` method public which I will do here as another commit.
  
> I think we could avoid allocating in the `json` and `form` case if we did. Also we'd permit people to stream arbitrary stuff into the request.

The streaming part will come as written above since my primary aim is to close #55 eventually.

But note that I tried making e.g. `json` non-buffering with zeptohttpc and this fails with various public API which do not accept chunked bodies and require a  `Content-Length` header. This is why I added `json_buffered` to zeptohttpc. But for attohttpc, I think we should go the other way around to avoid breaking existing code and being as easy to use as possible.

There is also a significant overhead for the chunked encoding with small JSON documents as `serde_json` will write really small chunks like `{` or `[` separately. This this does not impact network I/O overhead as there is a buffered writer below the chunked encoding layer, but the body size itself increases significantly. (We should probably add another layer of buffered writing directly below the JSON writer for that case if we enshrine it in API at all.)

adamreichold:
> User code should also already be able to implement the trait and the only thing left to make this fully work is to make the RequestBuilder::body method public which I will do here as another commit.

Added the commit.

sbstp:
Sorry for the delays, I'm quite busy lately. LGTM

adamreichold:
> Sorry for the delays, I'm quite busy lately. LGTM

Please don't feel pressured. I am late with the implementation itself as well. I guess this is spare time work for the both of us.

PR #61, opened 2020-04-24
Enable inspection of request builder properties before preparation

### LINKED ISSUE #60: Provide getters for RequestBuilder
I'd like to propose methods to be able to read fields from `RequestBuilder` before it is prepared (or sent). These would mirror the methods available in `PreparedRequest`:
```rust
fn method(&self) -> &Method;
fn headers(&self) -> &HeaderMap;
fn body(&self) -> &[u8];
```
~(and while these names don't clash with `RequestBuilder`'s current API, I'd personally prefix them as `get_method()`, `get_headers()`, and `get_body()`)~

The goal of these is to allow users to build their own extension traits over `RequestBuilder`. For example, I'm working on an interface that requires signing each request with a cryptographic key, and adding that signature to the request's headers. The most convenient way would be to do something like:
```rust
trait RequestBuilderExt { .. }

impl RequestBuilderExt for RequestBuilder {
    fn sign_message(self) -> Self {
	// Sign the message using some combination of the 
	// request's method, url, and body
	let signature = ..;
	
	// Add the signature header to the request
	self.header("ACCESS-SIGNATURE", signature);
	
	self
    }
}
```
which would let my users simply call `.sign_message()` on a `RequestBuilder` prior to sending it. 

As it stands right now, I can't read existing fields from a `RequestBuilder`, and I can't mutate a `PreparedRequest`.

Would these new APIs be welcome to attohttpc?

(and, thanks for writing attohttpc, of all http clients out there it has my favorite API)

adamreichold:
Closes #60

### DIFF
--- a/src/request/builder.rs
+++ b/src/request/builder.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use http::{
-    header::{HeaderValue, IntoHeaderName, ACCEPT, CONNECTION, CONTENT_LENGTH, USER_AGENT},
+    header::{HeaderMap, HeaderValue, IntoHeaderName, ACCEPT, CONNECTION, CONTENT_LENGTH, USER_AGENT},
     Method,
 };
 #[cfg(feature = "tls-rustls")]
@@ -22,8 +22,6 @@ use crate::request::{
     body::{self, Body, BodyKind},
     header_append, header_insert, header_insert_if_missing, BaseSettings, PreparedRequest,
 };
-
-const VERSION: &str = env!("CARGO_PKG_VERSION");
 
 /// `RequestBuilder` is the main way of building requests.
 ///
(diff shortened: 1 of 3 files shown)


sbstp:
LGTM

src/request/builder.rs
```diff
@@ -422,6 +420,39 @@ impl<B: Body> RequestBuilder<B> {
...
+impl<B> RequestBuilder<B> {
+    /// Inspect the properties of this request
+    pub fn inspect(&self) -> RequestInspector<'_, B> {
+        RequestInspector(self)
+    }
+}
+
+/// Allows to inspect the properties of a request before preparing it.
+#[derive(Debug)]
+pub struct RequestInspector<'a, B>(&'a RequestBuilder<B>);
```
sbstp, line 432: Should be it called `RequestBuilderInspector`? It's technically what it is, but it feels a bit Java-esque.

adamreichold: The type name will probably never be spelled out in user programs, so I have no problem with changing the name. However, I do concur that `RequestBuilderInspector` is the road towards `AbstractBaseClassImplFactory` and would personally keep the simpler name.

sbstp: Yeah ok, it's not a big deal.

PR #62, opened 2020-04-24
Add support for chunked transfer encoding used for request bodies

### LINKED ISSUE #55: Support streaming request bodies

adamreichold:
Closes #55

### DIFF
--- a/src/request/builder.rs
+++ b/src/request/builder.rs
@@ -7,7 +7,9 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use http::{
-    header::{HeaderMap, HeaderValue, IntoHeaderName, ACCEPT, CONNECTION, CONTENT_LENGTH, USER_AGENT},
+    header::{
+        HeaderMap, HeaderValue, IntoHeaderName, ACCEPT, CONNECTION, CONTENT_LENGTH, TRANSFER_ENCODING, USER_AGENT,
+    },
     Method,
 };
 #[cfg(feature = "tls-rustls")]
@@ -401,6 +403,9 @@ impl<B: Body> RequestBuilder<B> {
             BodyKind::Empty => (),
             BodyKind::KnownLength(len) => {
                 header_insert(&mut prepped.base_settings.headers, CONTENT_LENGTH, len)?;
+            }
+            BodyKind::Chunked => {
+                header_insert(&mut prepped.base_settings.headers, TRANSFER_ENCODING, "chunked")?;
             }
         }
 
--- a/src/request/body.rs
+++ b/src/request/body.rs
@@ -9,6 +9,8 @@ pub enum BodyKind {
     Empty,
     /// A request body with a known length
     KnownLength(u64),
+    /// A request body that is transferred using chunked encoding
+    Chunked,
 }
 
 /// A generic rewindable request body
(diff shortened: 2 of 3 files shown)


adamreichold:
@sbstp This is just the infrastructure to write chunked bodies via user code without any library type using it. Would you like me to add any predefined streaming type like JSON?

src/request/builder.rs
```diff
@@ -400,7 +400,7 @@ impl<B: Body> RequestBuilder<B> {
         header_insert(&mut prepped.base_settings.headers, CONNECTION, "close")?;
         prepped.set_compression()?;
         match prepped.body.kind()? {
-            BodyKind::Empty => (),
```
sbstp, line 403: Should we set the Transfer-Encoding header to chunked?

adamreichold: Definitely! Again, forgot this while porting. Thank god for code review...

sbstp:
> @sbstp This is just the infrastructure to write chunked bodies via user code without any library type using it. Would you like me to add any predefined streaming type like JSON?

I think we can wait a bit before doing this. We will need to add some sort of buffering for serializers since most of them write one byte at a time to the stream, we certainly should not be sending chunks of 1 byte.

src/request/builder.rs
adamreichold: Fixed.

adamreichold:
> We will need to add some sort of buffering for serializers since most of them write one byte at a time to the stream, we certainly should not be sending chunks of 1 byte.

In zeptohttpc, I just wired in a `BufWriter` between the serializer and the chunked writer to fix this, e.g. https://github.com/adamreichold/zeptohttpc/blob/2b4bf4507d6e06f439e48aa04c7d920d17fe4254/src/body_writer.rs#L113 (I don't this is a limitation even with the default buffer size of 8 kB as streaming only makes sense for larger bodies anyway.)

sbstp:
Simple enough. I would still prefer to do it in a separate pull request though.

sbstp:
LGTM

sbstp:
@adamreichold I had an idea. Perhaps we could offer a default implementation of the `Body::kind` method that defaults to chunked? That way you don't always have to figure out how to implement it.

adamreichold:
> I had an idea. Perhaps we could offer a default implementation of the `Body::kind` method that defaults to chunked? That way you don't always have to figure out how to implement it.

I am skeptical as this not a good default as the user needs to understand what chunking means to e.g. not fall into the one-character-at-a-time trap of typical serializers. I think providing a sensible set of predefined implementations might be a better usability improvement.

sbstp:
Yeah that's true, good point. I'm planning to release all these changes soon and we can improve the JSON and urlencoded bodies later, if that's ok with you.

adamreichold:
> Yeah that's true, good point. I'm planning to release all these changes soon and we can improve the JSON and urlencoded bodies later, if that's ok with you.

I don't think we can do much about URL-encoded forms as the only supported target is `String` but not `io::Write`, i.e. https://docs.rs/url/2.0.0/url/form_urlencoded/trait.Target.html

I will prepare one more MR for streaming large JSON documents, this rounds out the feature set and would make a good release candidate IMHO.

PR #65, opened 2020-04-24
Add built-in support for streaming JSON bodies.

adamreichold:


sbstp:
Do you think there's value in keeping the non streaming version?

adamreichold:
> Do you think there's value in keeping the non streaming version?

Yes, we definitely need it as some servers do not accept chunked request bodies at all. Especially for API that have strict limits on the request body size anyway and enforce the existence of a `Content-Length` header, c.f. https://github.com/sbstp/attohttpc/pull/59#issuecomment-616581963

sbstp:
Ok, fair enough then. I'll merge this and a cut a release. Thanks for doing all the work, lol.

PR #68, opened 2020-05-28
Optional multipart support

leo60228:
I used the `multipart` crate for this, which doesn't seem to have been updated in over a year. However, there aren't any outstanding issues, and I couldn't find any other synchronous implementation that's not tied to a specific HTTP stack.

### DIFF
--- a/src/multipart.rs
+++ b/src/multipart.rs
@@ -7,18 +7,16 @@ use std::io::{copy, prelude::*, Cursor, Error as IoError, Result as IoResult};
 
 /// A file to be uploaded as part of a multipart form.
 #[derive(Debug, Clone)]
-pub struct MultipartFile {
-    name: String,
-    file: Vec<u8>,
-    filename: Option<String>,
+pub struct MultipartFile<'key, 'data> {
+    name: &'key str,
+    file: &'data [u8],
+    filename: Option<&'key str>,
     mime: Option<Mime>,
 }
 
-impl MultipartFile {
+impl<'key, 'data> MultipartFile<'key, 'data> {
     /// Constructs a new `MultipartFile` from the name and contents.
-    pub fn new(name: impl AsRef<str>, file: impl AsRef<[u8]>) -> Self {
-        let name = name.as_ref().to_owned();
-        let file = file.as_ref().to_owned();
+    pub fn new(name: &'key str, file: &'data [u8]) -> Self {
         Self {
             name,
             file,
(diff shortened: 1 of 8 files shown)


src/multipart.rs
```diff
@@ -0,0 +1,123 @@
...
+            filename: Some(filename.as_ref().to_owned()),
+            ..self
+        }
+    }
+}
+
+/// A builder for creating a `Multipart` body.
+#[derive(Debug, Clone, Default)]
+pub struct MultipartBuilder {
+    text: Vec<(String, String)>,
```
sbstp, line 58: I wonder if it would be possible to avoid allocation here, at least until we build the `Multipart` struct.

```rust
pub struct MultipartFile<'key, 'data> {
    name: &'key str,
    file: &'data [u8],
    filename: Option<&'key str>,
    mime: Option<Mime>,
}

pub struct MultipartBuilder<'key, 'data> {
    text: Vec<(&'key str, &'data str)>,
    files: Vec<MultipartFile<'key, 'data>>,
}

sbstp:
Thanks that looks good. I wonder if we could add an integration test for this, just to make sure everything is good.

We already use rouille in the tests/ folder, just a matter of checking if there's multipart support.

leo60228:
I tried that, but I'm getting a deadlock in `rouille`/server-side `multipart`'s internals and I can't figure it out... https://gist.github.com/63540bc4ed7170e79c9d01598762b645

sbstp:
Alright I'll give it a spin and see if I can help

sbstp:
I also got the deadlock.

sbstp:
I think we should try using a different http server just to make sure the issue isn't on our end. Since it's a dev dependency we can use anything we want. I think warp might be a good option.

Warp has multipart support https://docs.rs/warp/0.2.3/warp/filters/multipart/index.html

sbstp:
I can do the test on my side, I'll let you know about the outcome.

sbstp:
Wow you figured out how to use warp much faster than me. I'm still struggling to get it to compile 😄.

I think it would be better to use a channel than to use panics to signal if the request was properly created. The test can wait on the channel to receive a response from the server and then assert the result is ok.

sbstp:
https://travis-ci.org/github/sbstp/attohttpc/jobs/694374690#L585 seems like there's an error here

sbstp:
Thanks a lot. I'll release this soon.

sbstp:
I released this in 0.14.0, thanks again.

Shnatsel:
README.md still says that multipart is not supported. Should it be updated?

sbstp:
Yeah, I'll do it

PR #74, opened 2020-06-28
Allow to add additional root certificates with native-tls

### LINKED ISSUE #71: Allow to add a root certificate for native-tls
Disabling the whole certificate verification (#38) is sometimes a way you don't want to follow, even for development purposes. A much safer alternative if to trust an additional root certificate. It would be nice to implement this option in `attohttpc`, which should be easy since `native-tls` already implements the [add_root_certificate](https://docs.rs/native-tls/0.2.4/native_tls/struct.TlsConnectorBuilder.html#method.add_root_certificate) method.

Please note that `rustls` users already have this option in `attohttpc` through the [client_config](https://docs.rs/attohttpc/0.14.0/attohttpc/struct.RequestBuilder.html#method.client_config) method since the [ClientConfig](https://docs.rs/rustls/0.17.0/rustls/struct.ClientConfig.html) object has a [RootCertStore](https://docs.rs/rustls/0.17.0/rustls/struct.RootCertStore.html) field.

breard-r:
Fix #71


src/streams.rs
```diff
@@ -91,10 +91,13 @@ impl BaseStream {
...
     #[cfg(feature = "tls")]
     fn connect_tls(host: Host<&str>, port: u16, info: &ConnectInfo) -> Result<BaseStream> {
-        let connector = TlsConnector::builder()
-            .danger_accept_invalid_certs(info.base_settings.accept_invalid_certs)
-            .danger_accept_invalid_hostnames(info.base_settings.accept_invalid_hostnames)
-            .build()?;
+        let mut connector_builder = TlsConnector::builder();
+        connector_builder.danger_accept_invalid_certs(info.base_settings.accept_invalid_certs);
+        connector_builder.danger_accept_invalid_hostnames(info.base_settings.accept_invalid_hostnames);
+        for cert in info.base_settings.root_certificates.iter() {
```
adamreichold, line 97: I think the `.iter()` call could be replaced by just doing `for cert in &info.base_settings.root_certificates`?

src/certificate.rs
```diff
@@ -0,0 +1,21 @@
+use std::fmt;
+
+#[derive(Clone)]
+pub struct Certificate {
+    pub inner_cert: native_tls::Certificate,
+}
+
+impl fmt::Debug for Certificate {
```
adamreichold, line 8: I am not sure if the debug output is worth the additional module and new type wrapper compared to e.g. just using the existing `SkipDebug` wrapper? Especially so since we start to expose this in our API as per the base settings member.

If we decide to keep it, I would suggest using a tuple type like `pub struct Certificate(pub native_tls::Certificate)` and dropping the `From` impl as we never need to convert this in generic context.

src/streams.rs
breard-r: The compiler doesn't seems to like it:

```
error[E0507]: cannot move out of `info.base_settings.root_certificates` which is behind a shared reference
  --> src/streams.rs:97:21
   |
97 |         for cert in info.base_settings.root_certificates {
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |                     |
   |                     move occurs because `info.base_settings.root_certificates` has type `std::vec::Vec<certificate::Certificate>`, which does not implement the `Copy` trait
   |                     help: consider iterating over a slice of the `Vec<_>`'s content: `&info.base_settings.root_certificates`
```

adamreichold: You need the `&` sigil in front to indicate that you want a reference and move out of the struct field, i.e. `in &info.base_settings.root_certificates` instead of `in info.base_settings.root_certificates`.

src/certificate.rs
breard-r: > Especially so since we start to expose this in our API as per the base settings member.

It's not exposed. The public `add_root_certificate` methods accepts a `native_tls::Certificate`.

adamreichold: > It's not exposed. The public add_root_certificate methods accepts a native_tls::Certificate.

Indeed. I was mistakenly thinking about the `pub root_certificates: Vec<Certificate>` field in `BaseSettings` but that type itself is not exposed in the API.

Personally, I would still prefer to just use `SkipDebug` there to keep this reasonably simple.

breard-r: Sure. I just didn't knew there was a `SkipDebug` trait in the first place. I've just amended the PR so it uses it instead of the new `Certificate` module/type.

adamreichold: Looks much nicer to me now! (Totally a nit, but I guess `SkipDebug<Vec<Certificate>>` would be nicer as it would avoid having to spell every time an item is added and the resulting output will contain only one ellipsis instead of one for each certificate. `SkipDebug` also has a similar `From` impl as `Certificate` had.)

breard-r: Indeed, that's better this way. PR updated.

adamreichold:
@sbstp Looks good to me.

sbstp:
Looks good to me as well.

sbstp:
I'll release this in a bit

sbstp:
Thank you, it's out in 0.15.0

PR #86, opened 2021-02-05
Replace per-connection timeout with full request deadline

### LINKED ISSUE #85: Timeout not honored (program hang?) on some websites
On 16 websites of the top million, `attohttpc` with 10-second connect timeout and 30-second full timeout does not complete the request in 60 seconds.

Tested using [this code](https://github.com/Shnatsel/rust-http-clients-smoke-test/blob/main/attohttpc-smoke-test/src/main.rs). Test tool output from all affected websites: [attohttpc-hangs.tar.gz](https://github.com/sbstp/attohttpc/files/5933498/attohttpc-hangs.tar.gz)

ureq [has the same issue](https://github.com/algesten/ureq/issues/312), and the author has noticed that some of the affected websites have very long redirect chains. It is possible that the full request timeout is reset on every HTTP redirection.

adamreichold:
Fixes #85

### DIFF
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -4,6 +4,7 @@ use std::io::{self, Read, Write};
 use std::net::{Shutdown, TcpStream};
 use std::sync::mpsc;
 use std::thread;
+use std::time::Instant;
 
 use url::{Host, Url};
 
@@ -18,6 +19,7 @@ pub struct ConnectInfo<'a> {
     pub url: &'a Url,
     pub proxy: Option<&'a Url>,
     pub base_settings: &'a BaseSettings,
+    pub deadline: Option<Instant>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


sbstp:
Sorry I totally forgot about this PR. I will review it right now.

src/streams.rs
```diff
@@ -107,16 +109,20 @@ impl BaseStream {
...
-            .map(|timeout| -> Result<mpsc::Sender<()>> {
+            .deadline
+            .map(|deadline| -> Result<mpsc::Sender<()>> {
                 let stream = stream.try_clone()?;
                 let (tx, rx) = mpsc::channel();
                 thread::spawn(move || {
-                    if let Err(mpsc::RecvTimeoutError::Timeout) = rx.recv_timeout(timeout) {
+                    let shutdown = match deadline.checked_duration_since(Instant::now()) {
+                        Some(timeout) => rx.recv_timeout(timeout) == Err(mpsc::RecvTimeoutError::Timeout),
+                        None => rx.try_recv() == Err(mpsc::TryRecvError::Empty),
```
sbstp, line 122: I'm trying to understand this line. If the deadline is expired and the channel is still connected, shutdown the stream, right?

sbstp:
LGTM!

src/streams.rs
adamreichold: Exactly, the other error is `mpsc::TryRecvError::Disconnected` which means the request is already done when we reach this line which might happen if the thread is scheduled in after a significant delay. (`Ok(_)` should never happen as we use only disconnection to signal this.)

sbstp: Sounds good

PR #88, opened 2021-02-06
RFC: Use a lossy conversion in ResponseReader::text_utf8

### LINKED ISSUE #87: Fetching google.com without 'charsets' feature fails; works in other clients
Fetching google.com without enabling `charsets` feature fails with error: `Io Error: stream did not contain valid UTF-8`

This is strange because this works fine in e.g. `ureq` without charsets support and encoding_rs dependency not being present in the tree at all.

Code used for tests: https://github.com/Shnatsel/rust-http-clients-smoke-test
In the repository both clients have encoding support enabled; I've manually disabled it prior to this test.

adamreichold:
Fixes #87

### DIFF
--- a/src/parsing/response.rs
+++ b/src/parsing/response.rs
@@ -149,6 +149,9 @@ impl Response {
     ///
     /// If the `charsets` feature is disabled, this method is the same as calling
     /// `text_utf8`.
+    ///
+    /// Note that both conversions are lossy, i.e. they will not raise errors when
+    /// invalid data is encountered but output replacement characters instead.
     #[inline]
     pub fn text(self) -> Result<String> {
         self.reader.text()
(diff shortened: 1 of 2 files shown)


Shnatsel:
I admit I have no idea what the correct behavior is here. It's possible that refusing to process invalid UTF-8 is the right thing do to, and the lossy conversion used by e.g. ureq is wrong. Are there any standard specifying this?

sbstp:
It seems like `reqwest` also uses a lossy conversion https://docs.rs/reqwest/0.11.0/reqwest/struct.Response.html#method.text

adamreichold:
We also decided for lossy conversion when using `encoding_rs` as `encoding_rs_io` defaults to this behaviour. So purely for reasons of consistency, the lossy versions seems preferable.

Issue #94, opened 2021-02-16
"InvalidResponse: invalid header" error on some websites


Shnatsel:
On some websites, e.g. http://opentrainer.ru, attohttpc fails with the following error:

> InvalidResponse: invalid header

Firefox and curl work fine.

32 websites out of the top million from [Feb 3 Tranco list](https://tranco-list.eu/list/3G6L) are affected.

Tested using [this code](https://github.com/Shnatsel/rust-http-clients-smoke-test/blob/main/attohttpc-smoke-test/src/main.rs). Test tool output from all affected websites: [atto-invalid-header.tar.gz](https://github.com/sbstp/attohttpc/files/5990836/atto-invalid-header.tar.gz)

adamreichold:
The linked site contains a multi-line continuation of a header value
```
Content-Security-Policy: script-src 'self' 'unsafe-inline' 'unsafe-eval' 
      https://staticxx.facebook.com 
      https://connect.facebook.net/ 
      https://tagmanager.google.com/ 
      https://www.googletagmanager.com/ 
      https://www.google-analytics.com/ 
      https://vk.com/ 
      https://mc.yandex.ru/ 
      https://top-fwz1.mail.ru/ 
      https://static.criteo.net/ 
      https://yastatic.net/s3/metrika/
      https://sslwidget.criteo.com/
      https://apis.google.com/
      https://cdn.ampproject.org/v0.js
      https://cdn.ampproject.org/v0/amp-sidebar-0.1.js
      https://cdn.ampproject.org/v0/amp-form-0.1.js
      https://cdn.ampproject.org/v0/amp-accordion-0.1.js;
    frame-src 'self' 'unsafe-eval' 'unsafe-inline'
      https://www.opentrainer.ru
      https://mc.webvisor.com 
      https://mc.webvisor.org 
      http://webvisor.com
      https://mc.yandex.ru 
      https://www.opentrainer.ru/trainer-api/school/singlegame/ 
      http://www.adobe.com/go/getflashplayer 
      https://www.facebook.com/tr/ 
      https://staticxx.facebook.com/connect/xd_arbiter/r/
      https://www.youtube.com/embed/ 
      https://i.ytimg.com/vi/
      https://vk.com/widget_community.php
      https://www.facebook.com/v2.6/plugins/
      https://www.youtube.com/subscribe_embed
      https://accounts.google.com/o/oauth2/postmessageRelay
      https://9215081.fls.doubleclick.net/;
[...]

PR #96, opened 2021-02-20
RFC: Drop invalid headers instead of raising an error

### LINKED ISSUE #92: "Http Error: invalid HTTP header name" on some websites
On some websites, e.g. http://hk-bbc.com, attohttpc fails with the following error:

> Http Error: invalid HTTP header name

Firefox and curl work fine.

96 websites out of the top million from [Feb 3 Tranco list](https://tranco-list.eu/list/3G6L) are affected.

Tested using [this code](https://github.com/Shnatsel/rust-http-clients-smoke-test/blob/main/attohttpc-smoke-test/src/main.rs). Test tool output from all affected websites: [atto-invalid-header-name.tar.gz](https://github.com/sbstp/attohttpc/files/5990742/atto-invalid-header-name.tar.gz)

adamreichold:
This follows Firefox and cURL to support broken sites, for example sites that contain PHP errors mixed into the header part of their HTTP response:
```
Server: Apache/2.2.3 (CentOS)
PHP Warning: PHP Startup: Unable to load dynamic library '/usr/local/php5/lib/php//usr/lib/php/modules/mcrypt.so' - /usr/local/php5/lib/php//usr/lib/php/modules/mcrypt.so: cannot open shared object file: No such file or directory in Unknown on line 0, PHP Startup: Unable to load dynamic library '/usr/local/php5/lib/php//usr/local/php5/lib/php/libmcrypt.so' - /usr/local/php5/lib/php//usr/local/php5/lib/php/libmcrypt.so: cannot open shared object file: No such file or directory in Unknown on line 0
X-Powered-By: PHP/5.2.4
```

Fixes #92

### DIFF
--- a/src/parsing/response.rs
+++ b/src/parsing/response.rs
@@ -55,10 +55,15 @@ where
         let header = trim_byte(b' ', &line[..col]);
         let value = trim_byte(b' ', &line[col + 1..]);
 
-        headers.append(
-            HeaderName::from_bytes(header).map_err(http::Error::from)?,
-            HeaderValue::from_bytes(value).map_err(http::Error::from)?,
-        );
+        let header = match HeaderName::from_bytes(header) {
+            Ok(val) => val,
+            Err(err) => {
+                warn!("Dropped invalid response header: {}", err);
+                continue;
+            }
+        };
+
+        headers.append(header, HeaderValue::from_bytes(value).map_err(http::Error::from)?);
     }
 
     Ok((status, headers))


sbstp:
Sounds good to me.

sbstp:
Whoops looks like I missed something. You try to parse the header name and drop it if it's invalid, but isn't the reported error about an invalid **value**?

adamreichold:
> Whoops looks like I missed something. You try to parse the header name and drop it if it's invalid, but isn't the reported error about an invalid value?

The original error reported (and reproduced) was:

> Http Error: invalid HTTP header name

I also checked that this originated in `HeaderName::from_bytes` and that this change fix it at least for http://hk-bbc.com

sbstp:
Oh sorry my bad. Indeed it's the space in the header name. For some reason I thought it was the really long header value.

adamreichold:
You might be thinking of #94 or #97 where the issue header values, specific line-folded header values.

PR #97, opened 2021-02-20
RFC: Handle line-folded headers

### LINKED ISSUE #94: "InvalidResponse: invalid header" error on some websites

adamreichold:
Split header lines only at CRLF boundary and replacing LF by SP in the resulting values.

Fixes #94

### DIFF
--- a/src/parsing/buffers.rs
+++ b/src/parsing/buffers.rs
@@ -18,6 +18,29 @@ where
     Ok(n)
 }
 
+pub fn read_line_strict<R>(reader: &mut BufReader<R>, buf: &mut Vec<u8>, max_buf_len: u64) -> io::Result<usize>
+where
+    R: Read,
+{
+    buf.clear();
+    let mut reader = reader.take(max_buf_len);
+    let mut n = 0;
+
+    loop {
+        let k = reader.read_until(b'\n', buf)?;
+        n += k;
+
+        if k == 0 || buf[buf.len() - 1] != b'\n' {
+            return Err(io::ErrorKind::UnexpectedEof.into());
+        }
+
+        if k >= 2 && buf[buf.len() - 2] == b'\r' && buf[buf.len() - 1] == b'\n' {
+            buf.truncate(buf.len() - 2);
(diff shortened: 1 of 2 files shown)


src/parsing/buffers.rs
```diff
@@ -18,6 +18,28 @@ where
     Ok(n)
 }
 
+pub fn read_line_strict<R>(reader: &mut BufReader<R>, buf: &mut Vec<u8>, max_buf_len: u64) -> io::Result<usize>
```
sbstp, line 21: Can you add a test below for this function?

sbstp: Also maybe a test in this file as well https://github.com/sbstp/attohttpc/blob/master/src/parsing/response.rs#L21 to make sure it meshes well with an otherwise properly formatted response.

sbstp:
I'm ok with merging this even if it's legacy stuff. The way you have found to implement it seems rather inexpensive (no alloc).

src/parsing/buffers.rs
adamreichold: Yes, I can do that. But I already found one issue with it: If the header contains CR that is not part of the line ending, the current code will fail, so I will probably have to refactor it considerably which will make it even more complicated. I am really not too enthused about it...

adamreichold: Added tests here and there...

sbstp:
Nice! You found a pretty clean way of handling this mess lol.

PR #99, opened 2021-02-20
Accept invalid UTF-8 for location headers.

### LINKED ISSUE #93: "missing or invalid location header" error on some websites
On some websites, e.g. http://futureuae.com, attohttpc fails with the following error:

> InvalidResponse: missing or invalid location header

Firefox and curl work fine.

96 websites out of the top million from [Feb 3 Tranco list](https://tranco-list.eu/list/3G6L) are affected.

Tested using [this code](https://github.com/Shnatsel/rust-http-clients-smoke-test/blob/main/attohttpc-smoke-test/src/main.rs). Test tool output from all affected websites: [atto-invalid-location-header.tar.gz](https://github.com/sbstp/attohttpc/files/5990781/atto-invalid-location-header.tar.gz)

adamreichold:
Fixes #93

### DIFF
--- a/src/request/mod.rs
+++ b/src/request/mod.rs
@@ -254,9 +254,10 @@ impl<B: Body> PreparedRequest<B> {
                 .headers()
                 .get(http::header::LOCATION)
                 .ok_or(InvalidResponseKind::LocationHeader)?;
-            let location = location.to_str().map_err(|_| InvalidResponseKind::LocationHeader)?;
 
-            url = self.base_redirect_url(location, &url)?;
+            let location = String::from_utf8_lossy(location.as_bytes());
+
+            url = self.base_redirect_url(&location, &url)?;
 
             debug!("redirected to {} giving url {}", location, url);
         }


src/request/mod.rs
```diff
@@ -254,9 +254,10 @@ impl<B: Body> PreparedRequest<B> {
                 .headers()
                 .get(http::header::LOCATION)
                 .ok_or(InvalidResponseKind::LocationHeader)?;
-            let location = location.to_str().map_err(|_| InvalidResponseKind::LocationHeader)?;
 
-            url = self.base_redirect_url(location, &url)?;
+            let location = String::from_utf8_lossy(location.as_bytes());
```
sbstp, line 258: What to you think of trying to do `.to_str()` first and if it fails then try the `from_utf8_lossy`? The result can be put in a `Cow`. Avoids allocation on the happy path.

sbstp: Hum, turns out `from_utf8_lossy` already returns a `Cow`. Nevermind that! Rust is great.

PR #103, opened 2021-03-24
Limit number of stored headers...

### LINKED ISSUE #101: Panic on large amount of headers
When the server sends a very large amount of headers, attohttpc panics with the following message:

> thread 'main' panicked at 'requested capacity too large', /home/shnatsel/.cargo/registry/src/github.com-1ecc6299db9ec823/http-0.2.3/src/header/map.rs:1550:9

Backtrace:
```
   0: std::panicking::begin_panic
   1: http::header::map::HeaderMap<T>::grow
   2: http::header::map::HeaderMap<T>::append
   3: attohttpc::parsing::response::parse_response
   4: attohttpc::request::PreparedRequest<B>::send
   5: attohttpc::request::builder::RequestBuilder<B>::send
   6: attohttpc_smoke_test::main
```

You can reproduce the issue by running the following shell script on Linux and connecting via attohttpc to `localhost:8080`

```bash
#!/bin/bash
(
    echo -e "HTTP/1.1 200 OK\r"
    i=0
    while true; do
        echo -e "Header$i: foobar\r"
        ((i=i+1))
    done
) | nc -l localhost 8080
```
Tested using [this code](https://github.com/Shnatsel/rust-http-clients-smoke-test/blob/main/attohttpc-smoke-test/src/main.rs) for attohttpc.

### LINKED ISSUE #102: Denial of service if the server sends the same header over and over with different values
attohttpc will use an unbounded amount of memory  if the server sends the same header over and over with different values. This can be used to cause a denial of service via memory exhaustion.

You can reproduce the issue by running the following shell script on Linux and connecting via attohttpc to `localhost:8080`

```bash
#!/bin/bash

(
    echo -e "HTTP/1.1 200 OK\r"
    i=0
    while true; do
        echo -e "Header: foobar$i\r"
        ((i=i+1))
    done
) | nc -l localhost 8080
```
Tested using [this code](https://github.com/Shnatsel/rust-http-clients-smoke-test/blob/main/attohttpc-smoke-test/src/main.rs) for attohttpc.

adamreichold:
…to avoid local resource exhaustion by malicious server.

Fixes #101
Fixes #102

@sbstp @Shnatsel Two design issue I see here
* Do we silently drop the headers but continue to parse them (with e.g. a request timeout kicking in eventually) or raise an error and fail the request? (Another option might be to log once that we started dropping them.)
* Do we keep this a hard-coded constant (similar to what we do for `MAX_LINE_LENGTH`) or make it an option user code can modify via the API?

### DIFF
--- a/src/parsing/body_reader.rs
+++ b/src/parsing/body_reader.rs
@@ -58,7 +58,7 @@ fn is_chunked(headers: &HeaderMap) -> bool {
 
 fn parse_content_length(val: &HeaderValue) -> Result<u64> {
     let val = val.to_str().map_err(|_| InvalidResponseKind::ContentLength)?;
-    let val: u64 = u64::from_str_radix(val, 10).map_err(|_| InvalidResponseKind::ContentLength)?;
+    let val = val.parse::<u64>().map_err(|_| InvalidResponseKind::ContentLength)?;
     Ok(val)
 }
 
--- a/src/parsing/response.rs
+++ b/src/parsing/response.rs
@@ -22,6 +22,7 @@ pub fn parse_response_head<R>(reader: &mut BufReader<R>) -> Result<(StatusCode, 
 where
     R: Read,
 {
+    const MAX_HEADERS: usize = 1024;
     const MAX_LINE_LEN: u64 = 16 * 1024;
 
     let mut line = Vec::new();
@@ -41,10 +42,13 @@ where
             .map_err(|_| InvalidResponseKind::StatusCode)?
     };
 
+    // headers
     loop {
         buffers::read_line_strict(reader, &mut line, MAX_LINE_LEN)?;
         if line.is_empty() {
             break;
+        } else if headers.len() == MAX_HEADERS {
+            return Err(InvalidResponseKind::Header.into());
         }
 
         let col = line
--- a/src/parsing/response.rs
+++ b/src/parsing/response.rs
@@ -18,11 +18,10 @@ use crate::{charsets::Charset, parsing::TextReader};
 #[cfg(feature = "json")]
 use serde::de::DeserializeOwned;
 
-pub fn parse_response_head<R>(reader: &mut BufReader<R>) -> Result<(StatusCode, HeaderMap)>
+pub fn parse_response_head<R>(reader: &mut BufReader<R>, max_headers: usize) -> Result<(StatusCode, HeaderMap)>
 where
     R: Read,
 {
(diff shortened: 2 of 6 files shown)


Shnatsel:
I believe this should raise an error rather than silently truncate the headers. Otherwise you can still DoS the client by feeding it headers forever; even if they're not stored, the connection will be open forever, and this will exhaust something down the line (thread pool or the like). Raising an error is what the rest of the HTTP clients do.

We can start with a hard-coded constant to get some protection in place, but ideally this should be configurable by the API client.

adamreichold:
> I believe this should raise an error rather than silently truncate the headers. Otherwise you can still DoS the client by feeding it headers forever; even if they're not stored, the connection will be open forever, and this will exhaust something down the line (thread pool or the like). Raising an error is what the rest of the HTTP clients do.

Changed this to raise an error instead of silently dropping headers. As an aside, these connections should not stay open forever if a request timeout was set.

> We can start with a hard-coded constant to get some protection in place, but ideally this should be configurable by the API client.

Extended the request builder to make this configurable similar to how one can change the maximum number of redirections now.

Shnatsel:
I haven't looked at the code, but the description of the behavior looks good. Thanks!

sbstp:
Hey sorry for the delay, I've been super busy. This looks good, although I would say that the default value of 1000 header is perhaps too high. We allow 16 KiB max for headers, to if we allow 1000 headers we get close to 16 MiB of memory. Seems like a lot.

I personally have never seen a request with more than 50 headers. I think 100 would be a safe default, I think hyper uses 128.

Let me know what you think.

sbstp:
Hyper uses 100 it seems, at least for HTTP/1.x https://github.com/hyperium/hyper/blob/68d4e4a3db91fb43f41a8c4fce1175ddb56816af/src/proto/h1/role.rs#L24

adamreichold:
> I personally have never seen a request with more than 50 headers. I think 100 would be a safe default, I think hyper uses 128.

Changed the default to 100. It started out with the very large value when this was a compile-time constant. But if anyone deliberately abuses headers this way, they can just adjust the setting now.

sbstp:
Neat. Just one last thing I spotted, the max_headers setting is on the RequestBuilder but not on the Session.

adamreichold:
> Just one last thing I spotted, the max_headers setting is on the RequestBuilder but not on the Session.

Sorry, forgot about that. Should fixed now.

sbstp:
Thanks!

PR #104, opened 2021-03-26
Don't parse an empty environment variable as a URL

samanpa:
This prevents us from logging an unnecessary warning when the http_proxy environment is empty.

CC @sbstp


sbstp:
Thanks

PR #105, opened 2021-05-22
Add cookie feature

kingtong:
Here is my attempt to bring or initiate cookie support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ readme = "README.md"
 repository = "https://github.com/sbstp/attohttpc"
 
 [dependencies]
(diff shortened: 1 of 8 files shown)


adamreichold (commented):
If we do integrate this into the HTTP client itself, shouldn't we also include parsing `Set-Cookie` response headers to populate the cookie jar? Otherwise this does not seem to add anything that an application using the crate could not do itself as it essentially sets a header.

kingtong:
Thanks for your feedback. I'll add the support for `Set-Cookie` to populate the cookie jar and rework the set cookie portion.

Issue #100, opened 2021-03-14
IP address support for tls-rust


gilescope:
At the moment when using the tls-rust back end it's checking that every host is a valid DNS, but IP addresses are not valid DNS names and thus it fails here:

https://github.com/sbstp/attohttpc/blob/147f7aef2fba7440e9c835feacfdffac4e418697/src/tls/rustls_impl.rs#L63

Presumably we need to skip dns resolution in this case?

gilescope:
This is an interesting divergence from reqwest. When they have `.danger_accept_invalid_certs(true)` set, they manage (somehow?) to be able to work with an IP address.

adamreichold:
Are you sure this works with reqwest?

https://github.com/seanmonstar/reqwest/blob/c666b293a139d1d9d681115d1f5090622a240835/src/connect.rs#L280

would suggest the same issue as described here?

adamreichold:
Using the current `master` branch of reqwest and the diff
```diff
diff --git a/examples/blocking.rs b/examples/blocking.rs
index a2b81c5..a1b25c6 100644
--- a/examples/blocking.rs
+++ b/examples/blocking.rs
@@ -6,7 +6,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     println!("GET https://www.rust-lang.org");
 
-    let mut res = reqwest::blocking::get("https://www.rust-lang.org/")?;
+    let mut res = reqwest::blocking::ClientBuilder::new().danger_accept_invalid_certs(true).build()?.get("https://140.82.121.3").send()?;
 
     println!("Status: {}", res.status());
     println!("Headers:\n{:?}", res.headers());
```
and running
```sh
cargo run --no-default-features --features "blocking rustls-tls" --example blocking
```
yields
```
Error: reqwest::Error { kind: Request, url: Url { scheme: "https", username: "", password: None, host: Some(Ipv4(140.82.121.3)), port: None, path: "/", query: None, fragment: None }, source: hyper::Error(Connect, Custom { kind: Other, error: "invalid dnsname" }) }
```
indeed. (The IP address is what github.com resolves to here.)

adamreichold:
The rustls documentation also suggests that a `DNSNameRef` is required, e.g. https://docs.rs/rustls/0.19.0/rustls/struct.ClientSession.html

PR #111, opened 2021-10-14
Add a new feature called `rustls` which includes the optional dependency and the additionally required crates.

### LINKED ISSUE #67: rustls vs tls-rustls feature flags
When turning on `rustls` instead of `tls-rustls`, I always get a `InvalidBaseUrl` error.

I realize that `rustls` is the name of the optional dependency.

It would be good to abort compilation in this situation with `cfg!` and `std::compile_error!`.

adamreichold:
This is actually not a breaking change after all since both the original `tls-rustls` and the `rustls` feature now coexist and both enable the `rustls-opt-dep`, `webpki` and `webpki-roots` optional dependencies.

Fixes #67 

Follow-up to #100

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ log = "0.4"
 mime = {version = "0.3", optional = true}
 multipart = {version = "0.18", optional = true}
 native-tls = {version = "0.2", optional = true}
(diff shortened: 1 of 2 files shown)


sbstp:
Just need a rebase and I'll merge this

adamreichold:
> Just need a rebase and I'll merge this

Pushed.

PR #114, opened 2022-04-20
fix(multipart): disable default features

lucasfernog:
The `multipart` crate is outdated and has several security vulnerabilites (you can check them running `cargo install cargo-audit && cargo audit`). This PR disables its default features so it doesn't pull outdated crates that uses vulnerable versions of chrono, time and hyper.


sbstp:
Thanks a lot.

lucasfernog:
Thank you!

PR #115, opened 2022-04-21
feat: allow changing the `Connection` header

lucasfernog:
The `Connection` header is always set to `close` even if we set it manually in the request builder. If the proposed solution is not desired, we could also expose a header method in the PreparedRequest struct.

Related to https://github.com/tauri-apps/tauri/issues/3889

### DIFF
--- a/src/request/builder.rs
+++ b/src/request/builder.rs
@@ -422,7 +422,7 @@ impl<B: Body> RequestBuilder<B> {
             base_settings: self.base_settings,
         };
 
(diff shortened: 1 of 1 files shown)


sbstp:
I don't see any issue with this. But note that attohttpc doesn't support connection pooling, so even if the users set `Connection: keep-alive` `attohttpc` will close the connection after reading the response.

lucasfernog:
Good to know, thanks! I'll just notify the user about this, so I think this PR isn't a priority.

PR #116, opened 2022-04-22
fix(multipart): always stream body

lucasfernog:
The multipart crate used by attohttpc has an issue when the form has only text fields, they are not computing the content length properly. I believe the problem is [here](https://github.com/abonander/multipart/blob/f4fee608af93bcad363b080849e4c3c56b622217/src/client/lazy.rs#L282) where the `text_data` is appended with the proper value but the `content_len` is not changed (so you only get the right body posted when `use_len === false`. I could open a PR to their repo but I'm afraid it won't be merged since the project seems unmaintained, so I open this change here hoping it won't cause any issues.

### DIFF
--- a/src/multipart.rs
+++ b/src/multipart.rs
@@ -96,10 +96,7 @@ pub struct Multipart<'data> {
 
 impl Body for Multipart<'_> {
     fn kind(&mut self) -> IoResult<BodyKind> {
-        match self.data.content_len() {
-            Some(len) => Ok(BodyKind::KnownLength(len)),
-            None => Ok(BodyKind::Chunked),
-        }
+        Ok(BodyKind::Chunked)
     }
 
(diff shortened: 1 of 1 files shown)


lucasfernog:
Forgot to mention, you can see the issue by removing the `with_file` call in the multipart example.

sbstp:
Ok that's sounds reasonable. Thanks for fixing this.

sbstp:
out in `v0.19.1`

lucasfernog:
Thank you :heart:

adamreichold:
Note this might be problematic: I stumbled over API that forbid chunked encoding of post bodies (presumably to simplify their body length checks). If upstream is really unmaintained, then perhaps searching an alternative or creating a fork is in order?

PR #119, opened 2022-05-26
Try implementing timeouts on Windows by directly closing the underlying socket from the timeout thread.

### LINKED ISSUE #118: Timeout not working on Windows
I'm making a request with timeout set to `Duration::from_secs(2)` to a dummy localhost server. The server sleeps for `Duration::from_secs(5)` before responding, and the timeout is not enforced on Windows (though it works on macOS and Linux).

adamreichold:
Closes #118

### DIFF
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -1,7 +1,14 @@
 #[cfg(test)]
 use std::io::Cursor;
 use std::io::{self, Read, Write};
-use std::net::{Shutdown, TcpStream};
+#[cfg(not(windows))]
+use std::net::Shutdown;
+use std::net::TcpStream;
+#[cfg(windows)]
+use std::os::{
+    raw::c_int,
+    windows::{io::AsRawSocket, raw::SOCKET},
+};
 use std::sync::mpsc;
 use std::thread;
 use std::time::Instant;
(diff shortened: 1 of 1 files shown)


adamreichold:
The test fails due to the wrong error condition being returned, but I'd like to know if this helps at all before trying to figure that one out.

lucasfernog:
Works like a charm @adamreichold

adamreichold:
> Works like a charm @adamreichold

Glad to hear that. I will look into fixing the error code next. Might take a day or two though.

adamreichold:
@sbstp I think this is ready for review now. The usage of unsafe is unfortunate, but I think unavoidable (as std does not call 'closesocket' itself) and mostly harmless (we just pass a socket handle, not raw pointers, into the unsafe function).

sbstp:
I wonder if this would be worth reporting to upstream, looks like it might be an issue/oversight in `std`.

Changes looks OK.

adamreichold:
> I wonder if this would be worth reporting to upstream, looks like it might be an issue/oversight in `std`.

I don't think so because std does not make any guarantees about the pattern we are using (closing/shutting down a socket blocked else to unblock that thread). It is just not part of their API contract and it working on Unix is a bit of coincidence.

PR #124, opened 2022-08-09
Add optional base64 dependency to make RequestBuilder::basic_auth generally available.

### LINKED ISSUE #123: Support 'basic_auth' on platforms other than Linux
Just want to start by saying thanks for building this crate - it's exactly what I need to make simple, synchronous http requests.

Question about the 'basic_auth' function of RequestBuilder - currently (v0.20.0) it is only supported on Linux:
https://docs.rs/attohttpc/latest/attohttpc/struct.RequestBuilder.html#method.basic_auth

Is there a reason for this limitation?  Can it be removed?  It's really just adding a header, so there doesn't seem to be any technical reason why this wouldn't work on all platforms.  (FWIW - for now I'm just adding my own 'basic_auth_creds' trait to RequestBuilder that adds this header, and it works just fine an all my target platforms - Linux, MacOS, Windows).

Thanks :)

adamreichold:
Tested this manually using the `cat` example against [httpbin.org](https://httpbin.org/#/Auth/get_basic_auth__user___passwd_) to ensure that the default base64 variant is the correct one. I also removed the part of the doc comment mentioning the required feature as docs.rs displays this itself nowadays.

Closes #123

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ readme = "README.md"
 repository = "https://github.com/sbstp/attohttpc"
 
 [dependencies]
+base64 = {version = "0.13", optional=true}
 encoding_rs = {version = "0.8", optional = true}
 encoding_rs_io = {version = "0.1", optional = true}
 flate2 = {version = "1.0", optional = true}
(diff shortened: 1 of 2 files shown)


sbstp:
Wikipedia says `The resulting string is encoded using a variant of Base64 (+/ and with padding).` This matches the STANDARD character set :+1:

PR #130, opened 2022-10-08
Allow to use flate2 with zlib or zlib-ng instead of miniz_oxide

jirutka:
This allows building flate2 against system-provided (or with vendored) zlib or zlib-ng library.

PR #131, opened 2022-10-09
Fix compress-zlib and compress-zlib-ng to actually enable compression

jirutka:
Follow-up for #130

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,12 +44,14 @@ tokio-stream = {version = "0.1.9", features = ["net"]}
 warp = "0.3.2"
 
 [features]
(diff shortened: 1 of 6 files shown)


sbstp:
Deployed as 0.23.1

adamreichold:
Sorry for being late to the party but I think it would have been simpler to qualify the code on

```rust
#[cfg(feature = "flate2")
```

as each optional dependency is already a feature itself instead of adding the pseudo-feature "_compress-any" which is not part of this crate's API if backwards compatibility is interpreted strictly.

jirutka:
> as each optional dependency is already a feature itself

I didn’t know that, thanks for the tip!

PR #134, opened 2022-11-03
Add feature to support rustls-native-certs root certificates

### LINKED ISSUE #133: Support for `rustls-native-certs`
Inspired by the `rustls-tls-native-roots` feature in [Reqwest](https://docs.rs/reqwest/latest/reqwest/#optional-features), would this project be open to a feature flag to enable pulling root certificates from the native store via [`rustls-native-certs`](https://crates.io/crates/rustls-native-certs/0.6.2)?

It's possible today to add them manually using the `add_root_certificate` method, however that adds a couple of complications: 

* The consumer needs to make sure to exactly match the `rustls-native-certs` version to the `rustls` version used transitively within `attohttpc`
* The consumer needs to create helpers to make sure that all requests go through a `RequestBuilder` that adds those certs.
* Some native stores hold old certs with invalid formats, which are rejected when they're added using `add_root_certificate` (see a comment in the [`reqwest` implementation](https://github.com/seanmonstar/reqwest/blob/231b18f83572836c674404b33cb1ca8b35ca3e36/src/async_impl/client.rs#L363-L365))

Having the ability to use the native store built in to attohttpc would alleviate all of these and allow consumers to opt-in easily via a feature flag.

I'm happy to take on creating a PR, if it's a feature you're open to adding.

charlespierce:
Closes #133 

Info
-----
* As mentioned in the issue, adding a feature flag to enable `rustls-native-certs` as root certificates allows consumers to easily opt-in to the native certificate stores.
* Inspired by how the similar feature works in Reqwest, the feature flag for this is all-inclusive, automatically activating `tls-rustls` as well.

Changes
-----
* Added a feature flag `tls-rustls-native-roots` to enable loading the native certificates as root certificates for requests.
* If the feature flag is enabled, load all of the native certs as root certificates.
    * We also ignore errors, as the native certs are sometimes invalid and we don't want to interrupt the entire request process if the user has an invalid cert somewhere in their native store.

Notes
-----
* I left the webpki certificate loading in regardless of the feature flag, as that's the current behavior and it seemed most reasonable to have the feature flag be additive, rather than a toggle.
* I'm not sure if we want to have a "valid" counter the way that Reqwest does, so that we can error out if the native cert check failed to load any valid certs? If that's a useful feature I'm happy to add, but I left it out in the interest of simplicity.
* I also updated a `"."` to `'.'` to resolve a clippy warning about using a single-character string as a pattern.

### DIFF
--- a/src/tls/rustls_impl.rs
+++ b/src/tls/rustls_impl.rs
@@ -9,6 +9,8 @@ use rustls::{
     client::{ServerCertVerified, ServerCertVerifier, WebPkiVerifier},
     ClientConfig, ClientConnection, OwnedTrustAnchor, RootCertStore, ServerName, StreamOwned,
 };
+#[cfg(feature = "tls-rustls-native-roots")]
+use rustls_native_certs::load_native_certs;
 use webpki_roots::TLS_SERVER_ROOTS;
 
 use crate::{Error, ErrorKind, Result};
@@ -55,6 +57,16 @@ impl TlsHandshaker {
                 root_store.add_server_trust_anchors(TLS_SERVER_ROOTS.0.iter().map(|root| {
                     OwnedTrustAnchor::from_subject_spki_name_constraints(root.subject, root.spki, root.name_constraints)
                 }));
+
+                #[cfg(feature = "tls-rustls-native-roots")]
+                for native_cert in load_native_certs()? {
+                    let cert = rustls::Certificate(native_cert.0);
+                    // Inspired by https://github.com/seanmonstar/reqwest/blob/231b18f83572836c674404b33cb1ca8b35ca3e36/src/async_impl/client.rs#L363-L365
+                    // Native certificate stores often include certificates with invalid formats,
+                    // but we don't want those invalid entries to invalidate the entire process of
+                    // loading native root certificates
+                    let _ = root_store.add(&cert);
+                }
 
                 for cert in &self.additional_certs {
                     root_store.add(cert)?;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ log = "0.4.17"
 mime = {version = "0.3.16", optional = true}
 multipart = {version = "0.18.0", default-features = false, features = ["client"], optional = true}
 native-tls = {version = "0.2.10", optional = true}
+rustls-native-certs = { version = "0.6", optional = true}
 rustls-opt-dep = {package = "rustls", version = "0.20.6", features = ["dangerous_configuration"], optional = true}
 serde = {version = "1.0.143", optional = true}
 serde_json = {version = "1.0.83", optional = true}
@@ -58,6 +59,7 @@ rustls = ["tls-rustls"]
 tls = ["native-tls"]
 tls-vendored = ["native-tls/vendored"]
 tls-rustls = ["rustls-opt-dep", "webpki", "webpki-roots"]
+tls-rustls-native-roots = ["tls-rustls", "rustls-native-certs"]
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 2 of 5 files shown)


src/tls/rustls_impl.rs
```diff
@@ -9,6 +9,8 @@ use rustls::{
     client::{ServerCertVerified, ServerCertVerifier, WebPkiVerifier},
     ClientConfig, ClientConnection, OwnedTrustAnchor, RootCertStore, ServerName, StreamOwned,
 };
+#[cfg(feature = "tls-rustls-native-roots")]
+use rustls_native_certs::load_native_certs;
 use webpki_roots::TLS_SERVER_ROOTS;
```
sbstp, line 14: Shouldn't `webpki_roots` be disabled when native certs are used?

sbstp: I think that it's likely that users that enable the native certs feature do not want the webpki certs. They want to rely on the OS' cert store and not on Mozilla's.

src/tls/rustls_impl.rs
```diff
@@ -56,6 +58,16 @@ impl TlsHandshaker {
...
                 }));
 
+                #[cfg(feature = "tls-rustls-native-roots")]
+                for native_cert in load_native_certs()? {
+                    let cert = rustls::Certificate(native_cert.0);
+                    // Inspired by https://github.com/seanmonstar/reqwest/blob/231b18f83572836c674404b33cb1ca8b35ca3e36/src/async_impl/client.rs#L363-L365
+                    // Native certificate stores often include certificates with invalid formats,
+                    // but we don't want those invalid entries to invalidate the entire process of
+                    // loading native root certificates
+                    let _ = root_store.add(&cert);
```
sbstp, line 68: We could maybe log an error using the `log` crate, this is also done if the proxy environment variables are not set properly.

charlespierce: Makes sense, I'll add that :+1:

src/tls/rustls_impl.rs
charlespierce: That's a good question that I wasn't totally sure of. I noted in the description that I left it as-is initially so that the feature flag would be additive as opposed to a toggle. Reqwest has two separate feature flags for webpki and native, so you could enable both or only one. I'm not super knowledgeable about the expected behaviors, so I don't have a good sense of if having only the native certs would cause issues.

charlespierce: Ah, that loaded weird, didn't see your second comment right away! But yeah, that makes sense. I'll update to make the native certs exclusive of the webpki ones.

sbstp: Doesn't have to be you who does this, but I think the TLS features should be cleaned up. Maybe something like:
* `tls-native`
* `tls-native-vendored`
* `tls-rustls-webpki-roots`
* `tls-rustls-native-roots`

charlespierce: Yeah, that would make sense. Would probably need to keep the various aliases around, at least for backwards compatibility, but it would definitely make it clearer which features are intended for which use-cases.

I took a quick look, my guess is it should be a separate PR, because there's a lot of various places that reference the current feature names that would need to be changed. I was also running into some weird behavior from Cargo that I didn't fully understand, where it wasn't importing the optional dependency even though it was declared to include it 🙁

charlespierce: Actually I think I got it working in #135 (which is stacked on this one to also include the `tls-rustls-native-roots` feature flag).

charlespierce:
@sbstp With this and #135 merged, would you be able to create / publish a new version on crates.io?

Not a super high priority, as we can use GH path imports if needed, but would be helpful when you have the time. Thanks in advance!

sbstp:
Yeah I'm planning to today

charlespierce:
Perfect, thanks!

sbstp:
Published 0.24.0

PR #135, opened 2022-11-03
Refactor TLS feature flags

charlespierce:
Info
-----
* Suggested in https://github.com/sbstp/attohttpc/pull/134#discussion_r1012520798, a cleanup of the various TLS feature flags so they are consistently named
* For compatibility, I kept the old names around as aliases, but removed the references to them in README and the root doc comments.

Changes
-----
* Updated the names of the core TLS features to be more consistent
* Added aliases to those with the existing feature names
* Added an internal feature `__rustls` to indicate Rustls support is enabled (with either kind of root certificates)
* Updated references in the code to the TLS features to use the new names

Notes
-----
* This PR builds on the changes in #134, so I'll leave it as a draft for now. To see only the new changes, check out only the [last commit](https://github.com/sbstp/attohttpc/pull/135/commits/0331de2b0927342d23d254bf4c50fc6a04f42d5d)

### DIFF
--- a/run-tests.bash
+++ b/run-tests.bash
@@ -22,5 +22,6 @@ cargo test --no-default-features --features form
 cargo test --no-default-features --features multipart-form
 cargo test --no-default-features --features json
 cargo test --no-default-features --features tls-native
+cargo test --no-default-features --features tls-native,tls-native-vendored
 cargo test --no-default-features --features tls-rustls-webpki-roots
 cargo test --no-default-features --features tls-rustls-native-roots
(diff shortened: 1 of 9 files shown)


run-tests.bash
```diff
@@ -21,5 +21,6 @@ cargo test --no-default-features --features compress-zlib-ng
 cargo test --no-default-features --features form
 cargo test --no-default-features --features multipart-form
 cargo test --no-default-features --features json
-cargo test --no-default-features --features tls
-cargo test --no-default-features --features tls-rustls
+cargo test --no-default-features --features tls-native
```
sbstp, line 24: Can you also test `tls-native-vendored`? Just so we know it builds properly

charlespierce: Good call, updated!

PR #137, opened 2023-02-11
Allow retrieving the final URL from the response

### LINKED ISSUE #136: Add a method on Response for retrieving the final URL
It would seem that there's no equivalent to *reqwest*'s [`url`](https://docs.rs/reqwest/latest/reqwest/blocking/struct.Response.html#method.url) method on the `Response` type?

baskerville:
Fixes #136.

### DIFF
--- a/src/parsing/response.rs
+++ b/src/parsing/response.rs
@@ -5,6 +5,7 @@ use http::{
     header::{HeaderName, HeaderValue, TRANSFER_ENCODING},
     HeaderMap, StatusCode,
 };
+use url::Url;
 
 use crate::error::{ErrorKind, InvalidResponseKind, Result};
 use crate::parsing::buffers::{self, trim_byte};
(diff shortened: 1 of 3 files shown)


src/parsing/response.rs
```diff
@@ -85,6 +86,7 @@ pub fn parse_response<B>(reader: BaseStream, request: &PreparedRequest<B>) -> Re
     headers.remove(TRANSFER_ENCODING);
 
     Ok(Response {
+        url: url.clone(),
```
sbstp, line 89: wouldn't `request.url()` be the same and avoid adding a param?

sbstp: nevermind, no, if there's a redirect it will be the wrong url

baskerville: Yes, I'll just need to update `self.url` in `send`'s loop.

sbstp: I think that what you are currently doing is better. I like to think of the PreparedRequest as immutable.

sbstp:
Thanks

baskerville:
If #138 gets merged, would you mind making a new release?

PR #145, opened 2023-08-23
Replace unmaintained webpki dependency with rustls-webpki.

macklin-10x:
Replaces the current dependency on the unmaintained webpki crate with rustls-webpki which seems to be a drop-in replacement.

Eliminates https://rustsec.org/advisories/RUSTSEC-2023-0052 from this crate.

I confess that I did not carefully understand how the current crate features fit together and made a very naive edit to them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ serde = {version = "1.0.143", optional = true}
 serde_json = {version = "1.0.83", optional = true}
 serde_urlencoded = {version = "0.7.1", optional = true}
 url = "2.2.2"
-webpki = {version = "0.22.0", optional = true}
+rustls-webpki = {version = "0.101.4", optional = true}
 webpki-roots = {version = "0.25.1", optional = true}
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -29,7 +29,7 @@ serde = {version = "1.0.143", optional = true}
 serde_json = {version = "1.0.83", optional = true}
 serde_urlencoded = {version = "0.7.1", optional = true}
 url = "2.2.2"
-webpki = {version = "0.22.0", optional = true}
+rustls-webpki = {version = "0.101.4", optional = true}
```
adamreichold, line 32: Is this dependency required at all? Doesn't the one on `webpki-roots` suffice?

macklin-10x:
Thanks for merging!

macklin-10x:
Is there a typical release cadence or do you plan to push a patch version soon?

sbstp:
I just published a new release

PR #153, opened 2024-01-20
Update to http 1

tottoto:
Update to `http` 1. As dev-dependencies require `http` 0.2 at this time, keeps to use it in tests where it is used.


sbstp:
Thanks

PR #155, opened 2024-01-21
Use rustls-pki-types via rustls crate

tottoto:
`rustls-pki-types` is only used in `__rustls` feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,7 @@ mime = {version = "0.3.16", optional = true}
 multipart = {version = "0.18.0", default-features = false, features = ["client"], optional = true}
 native-tls = {version = "0.2.10", optional = true}
 rustls-native-certs = { version = "0.7", optional = true}
-rustls-opt-dep = {package = "rustls", version = "0.22.0", optional = true}
-rustls-pki-types = {version = "1", optional = true}
+rustls-opt-dep = {package = "rustls", version = "0.22.1", optional = true}
 serde = {version = "1.0.143", optional = true}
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ multipart = {version = "0.18.0", default-features = false, features = ["client"]
 native-tls = {version = "0.2.10", optional = true}
 rustls-native-certs = { version = "0.7", optional = true}
 rustls-opt-dep = {package = "rustls", version = "0.22.0", optional = true}
-rustls-pki-types = "1"
+rustls-pki-types = {version = "1", optional = true}
```
adamreichold, line 28: I don't this dependency is necessary at all as the types are re-exported as `rustls::pki_types`.

tottoto: The re-exporting was introduced at `rustls` 0.22.1. So when we want to use `rustls-pki-types` crate via `rustls`, we have to set the `rustls` minimum version requirement as 0.22.1 instead of 0.22.0. Which option is appropriate for `attohttpc`?

adamreichold: Since 0.22.1 is fully semver-compatible with 0.22.0, I see little downside to bumping `rustls`. Most leaf projects will regularly run `cargo update` anyway so that the final result would be the same.

tottoto: Sure. I will change to use `rustls-pki-types` via `rustls` crate.

tottoto: Updated.

sbstp:
Awesome thanks to you both

PR #157, opened 2024-01-21
Apply rustfmt

tottoto:
This pull request proposes to format codes with `rustfmt`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
         with:
           toolchain: stable
       - run: rustup component add clippy
+      - run: cargo fmt --check
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/src/request/builder.rs
+++ b/src/request/builder.rs
@@ -135,7 +135,10 @@ impl<B> RequestBuilder<B> {
         };
         self.header(
             http::header::AUTHORIZATION,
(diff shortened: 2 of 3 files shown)


.github/workflows/.#ci.yml
```diff
@@ -0,0 +1 @@
+masa@masa-NUC12WSHi7.24580:1705762727
```
sbstp, line 1: What's this?

.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
         with:
           toolchain: stable
       - run: rustup component add clippy
+      - run: cargo fmt --all --check
```
adamreichold, line 34: This is a single crate, so why the `--all` flag?

.github/workflows/.#ci.yml
tottoto: Sorry, it is a mistake. Fixed.

.github/workflows/ci.yml
tottoto: Certainly. I will remove it.

tottoto: Fixed.

.github/workflows/.#ci.yml
tottoto: @sbstp Thank you for your reviews. Since this log contains information about my local development environment, would you mind if I ask you to delete the review comment so that it doesn't stand out? (I understand that force pushing do not remove the log, but I would like to make the log less noticeable as possible.)

PR #179, opened 2025-02-11
Move `tokio-rustls` to `[dev-dependencies]`

cuviper:
This was added to `[dependencies]` in #178, but I assume that was an accident because it's only used in tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,6 @@ rustls-opt-dep = { package = "rustls", version = "0.23.22", default-features = f
 serde = { version = "1.0.143", optional = true }
 serde_json = { version = "1.0.83", optional = true }
 serde_urlencoded = { version = "0.7.1", optional = true }
-tokio-rustls = "0.26.1"
 url = "2.2.2"
 webpki-roots = { version = "0.26.8", optional = true }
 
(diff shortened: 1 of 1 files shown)


cuviper:
Thanks for the quick update!

sbstp:
How did you notice the issue? It might be relevant to setup something to guard this from happening in the future.

cuviper:
I was checking `cargo update -n` (dry-run) on a project and noticed it wanted to add a **lot** of new dependencies. Then I tried my existing dependencies one by one until I saw that it was the `attohttpc` update pulling more in.

You could use [`cargo udeps`](https://crates.io/crates/cargo-udeps) as a safeguard, e.g.:

```
$ cargo +nightly udeps
    Checking attohttpc v0.28.4 (/home/jistone/rust/attohttpc)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.16s
info: Loading depinfo from "/home/jistone/rust/attohttpc/target/debug/deps/attohttpc-f7d55cb4a897d4d2.d"
unused dependencies:
`attohttpc v0.28.4 (/home/jistone/rust/attohttpc)`
└─── dependencies
     └─── "tokio-rustls"
Note: These dependencies might be used by other targets.
      To find dependencies that are not used by any target, enable `--all-targets`.
Note: They might be false-positive.
      For example, `cargo-udeps` cannot detect usage of crates that are only used in doc-tests.
      To ignore some dependencies, write `package.metadata.cargo-udeps.ignore` in Cargo.toml.
```

... and 0.28.5 is clean.

sbstp:
Awesome thanks, I'll add a CI step with udeps.

PR #180, opened 2025-02-13
Refactor rustls config

tottoto:
Refactors the rustls configs.


sbstp:
Ah that's great, thanks. I was wondering if we could use `dep:` to fix this mess.

PR #181, opened 2025-02-13
Remove unused rustls-pemfile crate from dev-dependencies

tottoto:
rustls-pemfile crate does not seem to be used.


sbstp:
Thanks, I think it was used before I refactored the tests to hyper 1.x, but it's no longer necessary since I use axum's TLS config object.

Issue #184, opened 2025-02-27
Feature: set Proxy-Authorization header if proxy url has user and password


ngicks:
Using a tool (i.e. volta) depending on attohttpc behind a proxy that requires authorization fails because the 407 status code.

PR #185, opened 2025-02-27
Add Proxy-Authorization header if proxy_url has username

ngicks:
quick fix for #184

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ readme = "README.md"
 repository = "https://github.com/sbstp/attohttpc"
 
 [dependencies]
-base64 = { version = "0.22.0", optional = true }
+base64 = { version = "0.22.0" }
 encoding_rs = { version = "0.8.31", optional = true }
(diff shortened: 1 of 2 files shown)


ngicks:
Tested with this build https://github.com/ngicks/volta/tree/feat-support-proxy-authorization
under my company's proxy, it is working.

If anything else is needed I would be happy to do that.

sbstp:
Thanks for the catch, this seems like a pretty big issue. I would probably enable this feature all the time, even if the basic-auth feature is not enabled, to avoid surprising behavior.

ngicks:
Force-pushed fixing casing (`basicAuth` -> `basic_auth`) to keeping commit history clean, I see MRs are merged without squash. 
Changed base64 to be non-optional. Proxy-Authorization head is always added if proxy_url has authorization info.

sbstp:
Yeah I use rebase merges when people take the time to make nice commits, squash otherwise.

sbstp:
Thanks

PR #190, opened 2025-06-14
Add feature flags `tls-rustls-webpki-roots-ring` and `tls-rustls-native-roots-ring` for rustls ring backend

metent:
I'm unable to build `attohttpc`-dependent source code into WASM components due to missing headers while compiling `aws-lc-sys`. Switching to `ring` backend fixes this issue.


sbstp:
aws-lc-rs remains the default backend for rustls. I don't mind adding a feature to use ring instead of aws-lc-rs, but I think the default should remain aws-lc-rs.

PR #191, opened 2025-06-14
Add a feature flag for disabling additional threads during initial connection and for timeout handling

metent:
This is useful for `wasm32-*` targets, which do not allow spawning additional threads.


sbstp:
Can you add an entry in `tools/test.bash` for this feature? Also the function feels quite harder to follow, it might be better to just make a different function for the case single threaded case.

PR #194, opened 2025-07-20
fix: remove deprecated `rustls` feature flag

metent:
When either of `tls-rustls-webpki-roots-ring` or `tls-rustls-native-roots-ring` are enabled, cargo pulls in `aws-lc-rs` too, presumably due to `__rustls-ring` depending on `rustls/*` features (and for some reason cargo assumes that we are also depending on `rustls` feature along with `rustls` dependency). I could not find any other way around it.


sbstp:
I've released v0.30.1 with this change, please let me know if it's ok now.

metent:
`awc-lc-rs` is no longer being pulled. Thanks.

PR #195, opened 2025-08-07
fix: update tls/mod.rs to use __rustls-ring feature

metent:
Realized that I hadn't pushed this fix upstream. Sorry for that!


sjackman:
I could use this feature `tls-rustls-webpki-roots-ring`, which needs this PR to work.

